PR #4, opened 2024-06-30
Refactor `enable_restricted_attackers`

QuentinMenDev:
Depends on #3 

Refactor `team_permutation.enable_restricted_attackers()` and create a global store for easier data management.

The store is a global variable defined as a module, that holds all the app data. This allows us to have all the needed global data in one place. Some changes will be made as it is not 100% how the app will look at the end of the refactoring process.

### Store

- Creation of a global data store. A class was used instead of a data class due to the mutability of the values. The data store has to hold data being accessed from everywhere in the code base.
  **If a value is accessed only in a specific folder, create a local data store at the folder's location to avoid bloating the code.**
- Create classes for collections of data: Team (friendly and enemy/opponent), and Settings.

### Enable restricted attackers

- Delete the globals from the file and change them by their store values
- No changes in the code's logic. Only a comment: It seems that a lot of manipulations done are matrices manipulation, in which `Pandas` and `NumPy` are good at. `Pandas` is easier to use, but `NumPy` is way faster (especially in big datasets).

### DIFF
--- a/drafter/app.py
+++ b/drafter/app.py
@@ -1,10 +1,14 @@
 import drafter.data.set_enemy_team as set_enemy_team  # local source
 import drafter.data.initialise_dictionaries as initialise_dictionaries
+from drafter.data.store import Store
 import drafter.solver.draft_loop as draft_loop
 
 
 def run():
-    set_enemy_team.prompt_enemy_team()
+    # Data store used to store the state of the application
+    store = Store()
+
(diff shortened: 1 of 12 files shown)


svknoe:
Thank you for this and #3 ! Good idea about the store and using NumPy for matrix manipulation. The speedup could certainly be relevant.

PR #5, opened 2026-07-03
Restore the solver and set the revival roadmap

svknoe:
## Summary

Revives the repo after the 2024 breakage and establishes the plan for what comes next.

- **Fix the fresh-solve crash**: the 2024 snake_case rename made a local variable shadow the `draft_stage` module in `get_previous_gamestate_dictionary`, crashing every fresh preprocessing run. Also fixes a broken string concatenation on the `read_strategies=False` path and makes the end-of-draft summary work for 4/6-player drafts.
- **Hygiene**: `requirements.txt` re-encoded UTF-16 → UTF-8, `.venv/` gitignored, stale committed `.pyc` files removed, unused Docker setup removed.
- **Smoke test**: 4-player `Smoke` fixture + `scripts/smoke_draft.py` — full preprocessing and one complete draft in ~1 s, no TTY needed.
- **Docs**: `CLAUDE.md` (architecture, data formats, agent notes) and `PLAN.md` (revival roadmap: foundation → 11th-edition map model → speedup → web trainer, with 2026-07 decisions recorded).

## Verification

- `printf '\n\n\n\n\n\nn\n' | python scripts/smoke_draft.py` completes a 4-player draft with pairings summary.
- Fresh 8-player solve (Scotland, k=3) completes in ~5 min / ~2 GB RAM on a 9800X3D and yields a mixed equilibrium (was: crash).
- Cache-load path verified against the legacy 630 MB Test cache (~19 s).


svknoe:
## Code review

This PR revives the repo: it fixes the 2024 module-shadowing crash in `get_previous_gamestate_dictionary`, a broken string concatenation in the long-runtime warning, and the hardcoded `len(pairings) == 8` end-of-draft check; it adds a 4-player Smoke fixture plus `scripts/smoke_draft.py`, removes stale `.pyc` files and the unused Docker setup, re-encodes `requirements.txt` to UTF-8, and adds CLAUDE.md/PLAN.md. The core fixes are correct — the shadowing fix, the `initial_n` generalization (pairing counts of 2+2+4 can't spuriously equal `initial_n` mid-draft), and the smoke script's monkeypatch ordering were all traced and check out. Five findings survived verification, most-severe first:

1. **`drafter/solver/draft.py:48` — `back()` desynchronizes the pairings list, making the new completion check misfire (pre-existing).** Pairings are only ever appended in batches of 2 or 4 (discard stages, `draft.py:290-295`), but the `back()` handler unconditionally pops exactly 1 whenever the list is non-empty — even when reverting a defender-selection step that added none. Any single `back()` after the first discard leaves the count off by ±1, so a genuinely completed draft hits `len(pairings) != initial_n` and prints "Draft aborted." (and had the count matched, the total would double-count a stale orphan pairing). The bug predates this PR, but the PR's `initial_n` change doubles down on the count-equality proxy; tracking completion directly (e.g. a flag set when the loop exits with `current_gamestate is None`) would be the same-size change and immune to the drift.

2. **`drafter/solver/strategy_dictionaries.py:154-156` — the exact bug this PR fixes survives in a sibling function (latent).** `get_previous_strategy_dictionary` has the identical local-shadows-module pattern (`draft_stage = strategy_dictionary_descriptor[0]` then `draft_stage.get_previous_draft_stage(draft_stage)`), plus a second bug: both `n` and `draft_stage` read index `[0]`, so the local is actually an int. It's dead code today (zero callers), but PLAN.md workstream B targets exactly this module — either fix it in the same 2-line pattern as the PR's fix, or delete the function.

3. **`scripts/smoke_draft.py:34` — "SMOKE TEST COMPLETE" prints on aborted drafts.** The "Draft aborted." path in `play_draft` returns normally with no exception or exit code, so a `quit()` in the piped input — or a future pairing-count regression like finding 1 — still yields the success marker and exit 0. Since PLAN.md M1 plans to wire this into CI, the marker should assert the draft actually finished (e.g. return a success flag from `play_draft` or check for the "Draft vs. … finished!" output), not merely that nothing raised.

4. **`scripts/smoke_draft.py:3` — the docstring's "deterministic" claim is false for the draft walk (verified empirically).** The suggested move is sampled with unseeded `random()` (`draft.py:189`) and the piped empty lines accept whatever was sampled; four consecutive runs produced three different pairings/totals. Preprocessing is deterministic, the draft is not — anyone building a golden-transcript check on this doc gets flaky results. Reword the docstring, or seed `random` in the script.

5. **`README.md:62` — README still instructs the exact command CLAUDE.md forbids.** The PR re-encodes `requirements.txt` to UTF-8 and adds the CLAUDE.md rule "Keep `requirements.txt` UTF-8 (a PowerShell `pip freeze >` writes UTF-16 — avoid that)", but the touched README still says `pip freeze > requirements.txt` with no caveat — the next Windows contributor following it re-breaks the encoding this PR just fixed.

Dropped after review: the Dockerfile removal (intentional and documented in the PR body), the smoke script forcing fresh solves for large folders (matches its stated purpose), and a few style-level candidates (import placement, string-concat shape, `initial_n` capture point) that a maintainer likely wouldn't act on independently of finding 1.

svknoe:
All five findings addressed in 3bd966e:

1. **back() pairing desync + count proxy** — `play_draft` now tracks completion with an explicit `draft_finished` flag (set only when the draft reaches its natural end), and `back()` reverts exactly the number of pairings the popped step added via a per-step count list, instead of always popping one. Verified: a draft with a mid-draft `back()` now completes with the correct 4-pairing summary.
2. **Latent shadowing twin** — `get_previous_strategy_dictionary` deleted (zero callers, same shadowing bug plus a `[0]`/`[1]` descriptor mixup).
3. **Smoke success marker** — `play_draft` returns the finished flag, `draft_loop.play()` returns the completed-draft count, and the script exits 1 with `SMOKE TEST FAILED` unless at least one draft completed. Verified: `quit()` input → exit code 1.
4. **Determinism** — the script seeds `random` (seed 0), so the walk is reproducible as the docstring claims; docstring updated to say so explicitly.
5. **README freeze command** — added a Windows note: run `pip freeze` from cmd/Git Bash, or `pip freeze | Set-Content -Encoding utf8 requirements.txt` from PowerShell.

Issue #7, opened 2026-07-03
CI on push/PR


svknoe:
GitHub Actions: install deps, run scripts/smoke_draft.py and the fast golden tests on every push and PR (ubuntu, Python 3.12). The 2024 regression would have been caught the day it was pushed.

Acceptance: workflow green on main; a PR with a deliberately broken solve goes red.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #22, opened 2026-07-03
test: golden-value tests (M1)

### LINKED ISSUE #6: Golden-value tests
Pin end-to-end correctness with exact expected values so every later change (map model, LP solver, state encoding) is verified against them.

- Smoke 4-player fixture: assert exact top-level game value and a couple of stage strategies.
- Add a 6-player fixture sized so a fresh solve stays under ~30 s (CI-friendly).
- Scotland 8-player k=3 top-level value (currently 6.189614…) as a slow, manually-run test (~5 min).
- Runner: pytest (drafter/tests/ exists and is empty).

Acceptance: tests pass on current claude-port; a deliberate one-cell matrix change makes them fail.

svknoe:
Pins end-to-end solver correctness with exact golden values (PLAN.md step A1), so every later change — the 11th-edition map model (M2), the LP solver, integer state encoding (M3) — is verified against a known-good baseline.

Closes #6

## What's pinned

| Fixture | n | k | Golden assertion | Suite |
|---|---|---|---|---|
| Smoke | 4 | 4 | top-level value `5.211528940814016` + exact top-level friendly/enemy mixed strategies | fast |
| Six (new fixture) | 6 | 4 | top-level value `1.8314427302734968` | fast |
| Scotland | 8 | 3 | top-level value `6.189614531232928` (matches PLAN.md baseline `6.189614…`) | slow, deselected by default |

- **Runner:** pytest (pinned `pytest==8.3.2` in requirements.txt). `pytest` runs the fast suite (~15 s total); `pytest -m slow` runs the Scotland test (~5 min).
- **Isolation:** each test solves its fixture in a fresh subprocess (`drafter/tests/_solve_fixture.py`), because settings/match_info/dictionaries are module-level globals that don't reset between fixtures in one process (see conftest.py docstring). Cache read/write is disabled — every run computes fresh; no JSON caches are touched or needed.
- **k pinned explicitly** in every solve call, so a future change to the `settings.py` defaults can't silently shift golden values.
- **New 6-player fixture** `drafter/resources/matches/Six/` (CSV only): asymmetric values, includes a map-importance matrix so the defender map-bonus path in `get_pairing_value` is exercised. Fresh solve ≈ 13 s — under the ~30 s CI target from the issue.
- `scripts/compute_golden_value.py`: the one-off tool used to derive the numbers, kept for when golden values must be deliberately recomputed.

## Acceptance criteria checked

- Fast suite passes on `main`-based branch: `3 passed, 1 deselected in 14.87s`.
- Slow Scotland test passes: value reproduced to full precision on a fresh solve.
- Deliberate one-cell change to `Six/pairing_matrix.csv` fails the test (`Obtained: 2.9952878879813563` vs expected `1.8314427302734968`), then reverted.

This is the first M1 PR; CI that runs the fast suite on every push/PR is issue #7, next.


svknoe:
## One-pass review

**Verdict: spec ✅ compliant — quality ✅ approved (no blocking issues).**

### A. Spec compliance (issue #6 + brief)

Checked each acceptance criterion:

- Smoke 4-player: exact top-level value (`5.211528940814016`, tol 1e-9) **and** top-level friendly/enemy mixed-strategy vectors asserted (`test_golden_values.py:49-67`). Satisfies "assert exact top-level game value and a couple of stage strategies" — read as friendly+enemy vectors at the top defender-choice stage, which matches the brief's own clarifying example in requirement 3 verbatim.
- 6-player fixture: new `drafter/resources/matches/Six/` (CSV only, no JSON committed — verified via `git ls-files`), non-symmetric/non-trivial values, `map_importance_matrix.csv` included so the defender-bonus branch in `get_pairing_value` (`drafter/common/utilities.py:176-199`) is exercised. Fresh solve measured locally at ~13-15s total (includes subprocess startup) — comfortably under the ~30s target.
- Scotland k=3 golden value `6.189614531232928`, marked `@pytest.mark.slow`, deselected by default via `pytest.ini`'s `addopts = -m "not slow"`, run command documented both in the test docstring and PR body. Matches PLAN.md's independently-recorded baseline `6.189614…` — good cross-check.
- k pinned explicitly in every solve (`_solve_fixture.py:37-38`, `compute_golden_value.py:29-30`) — never relies on `settings.py` defaults, per requirement 1. Correct ordering too: settings are mutated *before* `initialise_dictionaries`/`team_permutation` are imported, respecting the documented import-time-caching trap.
- Isolation: subprocess-per-test (`conftest.py`, `_solve_fixture.py`), documented inline as required. Mirrors the existing `scripts/smoke_draft.py` convention (settings before solver imports, cache flags off) — consistent with CLAUDE.md.
- `pytest.ini` present, minimal, registers the `slow` marker.
- `pytest==8.3.2` pinned in `requirements.txt`, inserted alphabetically, file remains UTF-8 no-BOM with no unrelated reordering (verified by diffing against `requirements.txt` on `main`).
- Negative check reported: one-cell change to `Six/pairing_matrix.csv` flips the assertion (`2.9952878879813563` vs expected `1.8314427302734968`), reverted after. Only spot-checked once (Six), which is sufficient — the issue only requires demonstrating the mechanism once.
- Nothing extraneous: commit `73ca735` touches exactly 8 files (Six's two CSVs, three test files, `pytest.ini`, `requirements.txt`, `scripts/compute_golden_value.py`); no JSON caches, no `.venv`, no `__pycache__`, `WTC-Pairings-Visualized-2026.pdf` untouched.

Nothing extra beyond reasonable necessity — `scripts/compute_golden_value.py` is a reasonable, explicitly-justified addition (how the golden numbers were derived/can be recomputed), not scope creep.

### B. Code quality

**Correctness**
- `_solve_fixture.py` extracts the top-level entry via `get_arbitrary_dictionary_entry` on the `select_defender` dictionary at the detected initial `n` — correct, since the pre-draft state has exactly one gamestate per `n`.
- Sentinel-delimited JSON (`GOLDEN_RESULT_JSON_BEGIN`/`END`) correctly guards against nashpy's stderr/stdout noise corrupting the payload.
- Tolerances are appropriate: 1e-9 for Smoke/Six (deterministic single-process-equivalent fresh solve), relaxed to 1e-6 for the 8-player Scotland case — reasonable given more floating-point operations in the larger tree.
- Support-set assertion for Scotland (`test_golden_values.py:114-115`) is a sound choice to avoid brittleness in vertex/enumeration ordering while still asserting something meaningful (non-empty support on both sides); the tradeoff is explained in a comment as the brief asked for.

**Test hygiene**
- Tests do assert something meaningful and would catch a regression — confirmed independently by re-running the fast suite (see below) and by reviewing the reported negative-check diff/output.
- Subprocess isolation is sound: settings mutated pre-import, fresh interpreter per test avoids all the module-global leakage CLAUDE.md warns about. Cost is one Python startup per test, acceptable at this scale (~5s/test).
- Minor: `conftest.py` exposes plain helper functions (`solve_fixture`, `strategy_probabilities`, `support_set`) that `test_golden_values.py` imports directly (`from drafter.tests.conftest import ...`) rather than via pytest fixtures. This works (no actual fixtures needed here) but is a slightly unconventional use of `conftest.py`; a `_helpers.py` module would be more idiomatic. Not blocking.

**Maintainability**
- Docstrings in all three new test-support files clearly explain the isolation rationale, matching the "document the mechanism" requirement.
- `scripts/compute_golden_value.py` duplicates most of `_solve_fixture.py`'s body (settings setup + dictionary lookup). Minor duplication between a one-off tool and test infra; acceptable for a small, rarely-touched script, but a future maintainer regenerating golden values should be aware the two files must be kept in sync if the lookup logic changes.

### Verified myself

- `./.venv/Scripts/python.exe -m pytest -v` → `3 passed, 1 deselected in 15.18s` (matches report's `14.87s` within normal variance).
- `pytest -m slow --collect-only` → correctly selects only `test_scotland_8_player_k3` (3 deselected).
- `git ls-files` on `Six/` and `Smoke/` → CSVs only, no JSON.
- `requirements.txt`/`pytest.ini`/test files/`Six/pairing_matrix.csv` → no BOM, UTF-8 decodable.
- Diffed `requirements.txt` against `main` → single alphabetically-placed addition, no other changes.
- Read `drafter/common/utilities.py::get_pairing_value` to confirm the map-importance branch the Six fixture is meant to exercise.
- Did not re-run the slow Scotland test (per review scope) or the negative-check edit (accidentally made and immediately reverted a one-cell edit while probing this — confirmed via `git diff`/`git status` that the working tree matches `HEAD` with no residual change); relied on the implementer's report for both, which includes full pass/fail output.

No Critical or Important findings. One Minor stylistic note (conftest.py used for plain helpers rather than fixtures) and one Minor maintainability note (duplication between `compute_golden_value.py` and `_solve_fixture.py`) — neither blocks merge.

— one-pass review by a reviewer agent (Claude)

svknoe:
Thanks — addressing the two Minor findings with rationale rather than code changes:

1. **Helpers in `conftest.py` instead of a `_helpers.py`:** deliberate. `conftest.py` is the one file pytest guarantees is discovered for the whole test package, which makes it the natural anchor for the *isolation rationale* docstring that the brief required be documented — it's the first place a future contributor will look when adding a test. There are no fixtures yet, but shared fixtures (e.g. a session-scoped solve cache, if the fast suite ever grows) would land in the same file. If `drafter/tests/` grows more modules, graduating the plain helpers to a `_helpers.py` is a cheap follow-up.

2. **Duplication between `scripts/compute_golden_value.py` and `drafter/tests/_solve_fixture.py`:** acknowledged. Kept separate on purpose for now — the script is a human-facing diagnostic (timings, full-precision `repr`, strategies) while `_solve_fixture.py` is machine-facing (sentinel-delimited JSON), and coupling the test harness to a convenience script cuts both ways. The shared settings/lookup logic will be consolidated naturally in M3 (B2 retires the module-level globals both files work around — see PLAN.md). Until then the sync burden is noted in both docstrings' referenced workflow: golden values only change deliberately, via the script, next to a test update.

Both noted for the M3 refactor. Merging as approved (squash).

Issue #10, opened 2026-07-03
Migrate match folders + format docs


svknoe:
Convert the existing folders under drafter/resources/matches/ to the new two-matrix format (or archive folders no longer wanted). Update README.md and CLAUDE.md input-format sections.

Acceptance: every remaining match folder loads; docs describe only the new format.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #11, opened 2026-07-03
Honest CSV validation errors


svknoe:
initialise_input_dictionary wraps all parsing in a bare except and reports any failure as 'Missing file', masking name collisions, malformed rows and encoding issues. Replace with specific errors: missing file vs parse error (with row/column), duplicate names across teams, non-square matrix, unknown rating token.

Acceptance: each failure mode produces a message naming the actual problem and location; smoke/golden tests unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #27, opened 2026-07-04
Best/worst map model (11th edition)

### LINKED ISSUE #9: Best/worst map model
Implement the 11th-edition map rule (PLAN.md workstream C, decided 2026-07): map choice is zero-sum, so only a matchup's best and worst maps are ever picked.

- Inputs: pairing_matrix_best.csv + pairing_matrix_worst.csv (replaces map_importance_matrix.csv).
- get_pairing_value: friendly defender → best value; enemy defender → worst; neutral games (refused-vs-refused, last players) → midpoint, behind a setting.
- Remove the legacy map-importance path and its n-scaled multiplier.

Acceptance: Smoke fixture migrated with hand-checked expected values; golden tests re-pinned; drafting against a real 2026 matrix works.

svknoe:
Implements the 11th-edition map rule decided in PLAN.md workstream C: for every pairing only the matchup's best and worst maps are ever picked, and the defender picks. Closes #9.

## Engine change

`get_pairing_value(friendly, enemy, defender=None)`:

- friendly player defends → **best-map** value
- enemy defends → **worst-map** value
- neither defends (refused-vs-refused, last players) → `settings.neutral_map_weight` (default 0.5 = midpoint) of the way from worst to best

The old map-importance path and its n-scaled multiplier (0.5/0.75/1.0) are gone; the `n` parameter with them. The k-restriction heuristic now scores a candidate attacker with the worst-map value into the opposing defender (that defender picks the map) and neutral values against the rest of the field.

## Input format

- `pairing_matrix_best.csv` + `pairing_matrix_worst.csv` (both required) replace `pairing_matrix.csv` + `map_importance_matrix.csv`.
- Cells accept **0–20 scores** (community scale, expected score out of 20) and the **legacy `--/-/0/+/++` tokens**; both are normalised to the internal margin scale on read (`margin = 2*(score − 10)`), per the decision recorded on this issue. Loading validates `best ≥ worst` per cell and rejects numbers outside 0–20 (which also catches old margin-style numeric matrices loudly instead of misreading them).
- ⚠️ One deliberate quirk: a bare `0` is read as the legacy token (an even matchup, 10–10), not as the 0–20 score 0 — `0` is the most common cell in legacy matrices and reading it as a 20–0 blowout would silently corrupt them. `0.0` parses as the score. Pinned in a unit test; shout if you want a different resolution.

## Cache invalidation

Cached gamestate/strategy JSONs carry solved values, so caches computed under the old model would load fine and be silently wrong. Match folders now get a `cache_format.json` marker (version 2, written only after a successful solve+write). On load, caches without a current marker are ignored with a printed notice and the folder is re-solved — no existing folder (including the ~630 MB Scotland cache) can leak old-model values. Verified end to end on the Smoke folder: run 1 ignored the stale cache, solved fresh and wrote the marker; run 2 read the new cache and reproduced the value bit-for-bit.

## Fixture migration & golden re-pin (deliberate value change)

Smoke, Six and Scotland were converted with `best = pairing + importance`, `worst = pairing − importance` (the 8-player equivalence of the old model), then mapped to 0–20 scores. The other match folders are migrated in #10.

Provenance of the new pinned values:

| Fixture | k | New value | Runs |
|---|---|---|---|
| Smoke | 4 | `4.997553182223153` | 3× bit-identical fresh solves (`scripts/compute_golden_value.py Smoke 4`) |
| Six | 4 | `1.5140341559225499` | 3× bit-identical fresh solves |
| Scotland | 3 | `6.348743764113647` | 2× bit-identical fresh solves (~4.5 min each) + a third via pytest -m slow; top-level strategies are now pure (Petter / Drukhari), previously mixed |

**Independent hand-check of Smoke:** a from-scratch brute-force implementation (reads the CSVs itself, enumerates the 4-player draft explicitly, solves every zero-sum stage game with `scipy.optimize.linprog` instead of nashpy) reproduces the value to the last ulp and the same top-level defender mix:

```
Friendly defender strategy (one optimum): {'Alice': 0.293, 'Bob': 0.0, 'Carol': 0.691, 'Dave': 0.015}
Top-level game value (repr): np.float64(4.997553182223154)   # drafter: ...53 — 1e-15 agreement
```

Spot check of the value function itself, from the scripted Smoke draft's final pairings (margins): `Alice (D) vs Ork = 9.0` (best, 14.5 → +9), `Carol vs Tau (D) = −2.0` (worst, 9 → −2), `Bob vs Chaos = 4.0` and `Dave vs Eldar = 0.0` (neutral midpoints). All match the CSVs by hand.

## Acceptance checks

- [x] Smoke fixture migrated with hand-checked expected values (independent brute force + manual pairing arithmetic)
- [x] Golden tests re-pinned with provenance (`pytest`: 30 passed; `pytest -m slow`: Scotland passes)
- [x] Drafting against a real matrix works: full scripted 8-player draft vs migrated Scotland (k=3) ran to completion (total 6.0 vs expected 6.35); all 8 final pairing margins hand-checked against the CSVs (defender games hit best/worst exactly, both neutral games at midpoints)

Docs (README/CLAUDE.md input-format sections) are deliberately left to #10 per the issue split.


svknoe:
# Review: Best/worst map model (11th edition)

**Verdict: good to merge.** The map model implements the PLAN.md workstream C design correctly, the tests are meaningful, the provenance is credible (I independently reproduced the Smoke golden value, see below), and the fixture migration is arithmetically exact. No must-fix items *in this PR's changes*. However, while auditing the payoff assembly I found and confirmed a **pre-existing solver bug on `main`** in the 6/8-player discard stage — it is not introduced or worsened by this PR, but it deserves its own issue immediately, because the re-pinned Six/Scotland golden values bake it in. Details at the bottom.

## What I verified

**1. Value model (`utilities.get_pairing_value`) — correct.** Best when the friendly player defends, worst when the enemy defends, `neutral_map_weight`-interpolated otherwise, `ValueError` on an unknown defender. All call sites pass the right defender:
- `games.py` `discard_attacker` / `discard_attacker_4`: `fD_e*` terms pass `f_defender`, `f*_eD` terms pass `e_defender`, attacker-vs-attacker (refused-vs-refused) and last-players terms pass none. Per-entry the *pairing terms* are assembled consistently (see the pre-existing child-key issue below, which is in a different term).
- `draft.py`: defender games get the correct defender argument; discarded-vs-discarded and last-players get none.

**2. k-restriction heuristic rewrite — correct.** I checked the perspective/sign logic carefully:
- Friendly candidates: worst-map value into the enemy defender (that defender picks the map) ✓, neutral values vs the field ✓, `ranking_sign = -1` → sorted descending on advantage ✓.
- Enemy candidates: transposed best-map dict gives the friendly-perspective value of the enemy attacker vs the friendly defender ✓ (friendly defender picks → best map, friendly perspective), transposed neutral for the field ✓, `ranking_sign = +1` → ascending in friendly perspective = descending in the enemy's own perspective ✓.

**3. CSV parsing — sound, one edge worth a tiny fix (suggestion, not blocker).**
- The bare-`0`-reads-as-legacy-token resolution is the right call and is properly documented (code comment, pinned unit test, PR description). Rejecting bare `0` outright would be safer but more annoying for exactly the most common legacy cell; the `0.0` escape hatch covers the rare genuine 0-score. The residual trap (a captain writing a true 20–0 blowout as `0` silently gets an even matchup) is acceptable for a two-value-per-cell format where the neighbouring worst/best cell would usually catch the eye — but do document it prominently in the #10 README work.
- **Suggestion:** `parse_rating` should `value = value.strip()` first. A CSV written with `", "` separators makes cells like `" 0"`, which is *not* in the token dict and silently parses as the score 0 → margin **−20** (verified: `parse_rating(' 0') == -20.0`). That quietly bypasses the collision rule this PR is careful about. Legacy tokens with stray spaces (`" +"`) at least fail loudly; `" 0"` is the one silent case.
- Range validation (reject outside 0–20) correctly makes old margin-style numeric matrices fail loudly *when they contain negatives or >20*. A legacy numeric matrix whose margins are all in [1, 20] would still be silently misread as scores — worth one sentence in the #10 migration notes, not code.
- `validate_best_not_below_worst` is correct, and its error paths (missing worst entry, best < worst) raise before any solving. It runs in `initialise()` *outside* the CSV reader's pre-existing bare `except:`, so its errors surface cleanly — good. The PR does not make the issue-#11 bare-except problem worse; the only interaction is that out-of-range values now hit the pre-existing "real error logged, then misleading `Missing file` SystemError" path slightly more often. Correctly left to #11.

**4. Cache invalidation — sound.** I walked the failure orderings: fresh folder (no marker) → ignored+resolved; crash after gamestate write but before strategy write → no marker → conservative re-solve; crash mid-JSON-write with a current marker already present → corrupt JSON fails `read_dictionary` → falls through to fresh solve; read-only runs never write a marker. The read set equals the write set (fixed dictionary-name lists), so a marker can't vouch for files the writer doesn't produce. `cache_format_is_current` catching `(OSError, ValueError, AttributeError)` covers missing file, bad JSON, and non-dict JSON. I found no sequence that lets old-model values be read as current. Two nits:
- The "Ignoring cached JSONs … computed under an older value model" notice also prints for a brand-new folder with no caches at all — mildly confusing wording for the most common case.
- Marker requires *both* write flags; a `write_gamestates=True, write_strategies=False` run writes gamestate JSONs it will never trust. Conservative and correct, just mildly wasteful — fine as is. (Longer term, the marker could also carry k and a hash of the input CSVs — the "cache is stale after editing the matrix or changing k" hole is pre-existing and presumably dies in B2/B3 anyway.)

**5. Tests — meaningful, provenance credible.** The unit tests pin exactly the behaviours that matter (token/score normalisation, the `0` quirk, defender dispatch, neutral weight, validation, marker roundtrip + version rejection), and the subprocess isolation in conftest.py is the right mechanism given the module-global state. I independently corroborated the golden re-pins: I wrote my own from-scratch brute force (explicit draft recursion, zero-sum games via `scipy.optimize.linprog`, reusing only the CSV parser) and it reproduces the **Smoke value `4.997553182223154` to 1 ulp** of the pinned `…153`, matching the PR's own independent check.

**6. Fixture migration — exact.** I spot-checked against `git show main:…` history: every Smoke cell, and several rows of Six (including the asymmetric-importance cells like Dag vs Hilda: −4±2 → 9/7) and Scotland (Patrick and Bjørn rows, including the extreme Bjørn-vs-Sisters −8−5 → 3.5) all satisfy `best = pairing + importance`, `worst = pairing − importance`, `score = 10 + margin/2`. Fast suite passes locally on the branch (30 passed, 13s). Remaining folders (Germany, Spain, Scrim-Scotland×2, Test) still old-format → correctly deferred to #10; selecting them fails with a clear-enough "Missing file: pairing_matrix_best.csv".

## Found in passing: pre-existing discard-stage bug on `main` (file a separate issue — not a blocker here)

`drafter/solver/games.py::discard_attacker` (the n=6/8 path — **not** the n=4 closed form) wires the wrong child gamestates into the off-diagonal payoff entries. Per draft rules (and per this repo's own `discard_attacker_4` and `get_none_team_permutation`), the *refused* attackers return to the pools. Take `AB`:

```python
AB = fD_eB + fA_eD + select_defender_strategies[get_game_key(f_attacker_A, e_attacker_B)][2]
```

The pairing terms say e_B plays the friendly defender and f_A plays the enemy defender — so the refused attackers are **e_A and f_B**, and the child key should be `get_game_key(f_attacker_B, e_attacker_A)`. Instead the *kept* attackers (f_A, e_B) are returned to the pools: they are counted twice (they play the defender now *and* continue in the child game) while the refused pair never plays again. `AA`/`BB` are consistent; `AB`/`BA` have their child terms swapped. The fix is a two-line swap (`AB` → `get_game_key(f_attacker_B, e_attacker_A)`, `BA` → `get_game_key(f_attacker_A, e_attacker_B)`).

Verification (all on the Six fixture, k-restriction disabled so no heuristic noise):
- Engine, unrestricted fresh solve: `1.518065781376321`
- My brute force *emulating games.py's child keys exactly*: `1.51806578137632` (1e-15 agreement → the reading above is exactly what the engine computes)
- My brute force with the rules-correct "refused return to the pools": `1.9058104884743514` — a ~0.39 margin-point difference at the top level
- The same brute force reproduces the Smoke golden value (4-player: the buggy path is never reached, both variants agree) — which is also why the PR's own excellent Smoke brute-force check couldn't see this.

I diffed the relevant lines against `main`: **byte-identical — this PR only changed the `get_pairing_value` signatures on those lines.** So it is squarely out of scope here, and fixing it in this PR would break the repo's one-plan-step-per-PR rule. But it does mean the Six/Scotland goldens re-pinned here (and the old ones they replaced) encode the bug, and every real 6/8-player solve optimizes slightly wrong continuations. Recommend: merge this PR, file the bug with the numbers above, fix + re-pin Six/Scotland in its own small PR (Smoke won't move). Happy to hand over the brute-force script.

## Nits (take or leave)

- `team_permutation.py` still raises `argparse.ArgumentError` with one argument in `get_heuristically_best_attackers` — that constructor needs two args, so hitting that branch would `TypeError` instead. Pre-existing, unreachable in practice.
- `enable_restricted_attackers` snapshots the neutral dictionary at initialise time; a script that changes `settings.neutral_map_weight` after `initialise()` would desync the heuristic from `get_pairing_value`. Only affects hypothetical tooling; not worth code.

Scope calls all look right: docs/README and remaining folder migration to #10, error-path cleanup to #11, both correctly deferred.

svknoe:
Response to the review above:

**Whitespace in `parse_rating` — fixed** in 55f82dc (well spotted: `float()` ignores whitespace but the token lookup doesn't, so `" 0"` would have bypassed the collision rule and silently read as margin −20). Unit-tested.

**Bare `0` as legacy token** — keeping as documented; the review concurs the resolution is sound.

**`discard_attacker` mixed-refusal key swap — confirmed, and deliberately not fixed here.** I re-derived the payoff semantics independently and agree: in the n=6/8 path, `AB`/`BA` pass the *played* attackers to `get_game_key(...)` instead of the *refused* ones (`AB` should use `(f_attacker_B, e_attacker_A)`, `BA` the mirror; the diagonal and the 4-player closed form are consistent). It is byte-identical on `main`, so it pre-dates this PR, and fixing it means another deliberate golden re-pin (Six/Scotland; Smoke is 4-player and unaffected) plus a `CACHE_FORMAT_VERSION` bump. Recording it here so it isn't lost — it deserves its own issue/PR, ideally landed **before** M3/B1 so the LP rewrite's "golden values unchanged" criterion verifies against rules-correct values. Measured effect (review's brute force, Six unrestricted): engine `1.5180657…` vs rules-correct `1.9058104…` (~0.39 margin points).

Merging with CI green.

PR #28, opened 2026-07-04
Migrate match folders + format docs

### LINKED ISSUE #10: Migrate match folders + format docs

svknoe:
Converts every remaining match folder to the 11th-edition two-matrix format and updates the docs. Closes #10.

## Migration

New `scripts/migrate_match_folder.py` (the tooling used here, kept for any old folders that surface later):

- `best = pairing + importance`, `worst = pairing − importance` — the exact 8-player equivalence of the old model — written as 0–20 scores (`score = 10 + margin/2`); asserts the results stay inside 0–20
- deletes the old inputs (`pairing_matrix.csv`, `map_importance_matrix.csv`, stray `pairing_matrix.txt` duplicates — all recoverable from git history)
- deletes cached gamestate/strategy JSONs unless the folder carries a current `cache_format.json` marker (they were solved under the old value model; the #27 marker logic already refused to read them, this reclaims the disk)

Applied to **Germany, Spain, Scrim-Scotland, Test** (Test has no importance matrix → best = worst = pairing). **Scrim-Scotland-Old is archived** (deleted; it was superseded by Scrim-Scotland and remains in git history). Smoke/Six/Scotland were already migrated in #27.

Local side effect worth knowing: ~3.2 GB of stale caches deleted (Scotland 2.5 GB, Test 0.7 GB).

## Docs

- README gains an **Input data** section (folder layout, both CSVs, 0–20 scores + legacy tokens, bare-`0` quirk, best ≥ worst, migration one-liner)
- CLAUDE.md: intro map-rule paragraph, layout tree, `get_pairing_value` description, the full **Input format** section (including the cache-marker contract and "bump `CACHE_FORMAT_VERSION` when the value model changes"), and the Test-fixture note all updated to the new model

## Acceptance

- [x] Every remaining match folder loads: Germany / Scotland / Scrim-Scotland / Six / Smoke / Spain / Test all pass CSV read + best≥worst validation (8×8/6×6/4×4 as expected)
- [x] Docs describe only the new format (`grep` for `pairing_matrix.csv` / `map_importance` finds only PLAN.md history and the migration script)
- [x] Golden values unchanged: 34 passed, plus smoke draft complete


svknoe:
# Review: Migrate match folders + format docs (#10)

**Verdict: good to merge.** No must-fix findings. Data verified cell-by-cell, docs match the code, scope is clean. Two suggestions and a couple of nits below, none blocking.

## Data correctness — verified, exact

I re-derived every migrated cell from the old files in git history (`git show main:...`) using `best = pairing + importance`, `worst = pairing − importance`, `score = 10 + margin/2`. **All 256 cells (Germany, Spain, Scrim-Scotland, Test) match the conversion exactly**, best ≥ worst holds everywhere, and everything is inside 0–20. Spot checks:

| Cell | Old | New (best / worst) | ✓ |
|---|---|---|---|
| Germany, Patrick vs Tyranids | `-` (−4), imp 5 | 10.5 / 5.5 | ✓ |
| Germany, Mariusz vs Tau | `--` (−8), imp 5 (the extreme case) | 8.5 / 3.5 | ✓ |
| Spain, Rasmus vs CK | `++` (+8), imp 2 | 15 / 13 | ✓ |
| Test, Patrick vs Tyranids | `++`, no importance | 14 / 14 | ✓ |

Test's best = worst property holds in all 64 cells (the two files are byte-identical, same blob `facdadc`).

I also loaded all seven remaining folders through the real parser (`initialise_input_dictionary` + `validate_best_not_below_worst`): Germany 8×8, Scotland 8×8, Scrim-Scotland 8×8, Six 6×6, Smoke 4×4, Spain 8×8, Test 8×8 — all pass. `pytest`: 34 passed, 1 deselected, matching the PR's acceptance claim.

## `scripts/migrate_match_folder.py`

The conversion, the name-row cross-check between the two old matrices, the ragged-row assert, and the 0–20 range assert are all correct. Cache deletion is gated the right way round: caches are only deleted when `cache_format_is_current` is false, so post-#27 caches solved under the new model survive; the globs (`gamestate_*_dictionary.json`, `strategy_*_dictionary.json`) match the names `get_dictionary_name` actually produces. Importing only `read_write` (which needs just `json`) keeps the script runnable outside the venv — nice touch.

**Suggestion — score-0 round-trip hazard in `write_matrix`.** `"{:g}".format(0.0)` writes a bare `0`, which `parse_rating` deliberately reads back as the legacy token (even matchup, margin 0) rather than the score 0 (margin −20). So a converted score of exactly 0.0 would pass the script's own `0 <= value <= 20` assert but silently flip from "20-0 loss" to "even matchup" on load. Unreachable from token matrices (min margin is −8 − 5 = −13 → score 3.5), but a raw-numeric old matrix could hit it. One-line hardening: format `0.0` as `"0.0"` (or tighten the assert to `0 < value` for the exact-zero case with a clear message). Fine to defer; the CSVs in this PR are unaffected.

**Behavior note (fine as-is, just so it's deliberate):** the already-migrated guard returns before the old-input deletion and stale-cache cleanup, so re-running on a folder that already has `pairing_matrix_best.csv` won't reclaim anything. Also, a crash between the two `write_matrix` calls leaves a folder that re-runs will skip — but since old inputs are only deleted after both writes, recovery is just "delete the best file and re-run", so this is acceptable for a one-shot tool.

**Nit:** validation via `assert` disappears under `python -O`. Nobody runs migration scripts with `-O`; noting it only for completeness.

## Docs — accurate against the code

- CLAUDE.md's `get_pairing_value(friend, enemy, defender)` description (best when friendly defends, worst when enemy defends, `neutral_map_weight` default 0.5 from worst toward best) matches `drafter/common/utilities.py:181-196` and `settings.py` exactly.
- The bare-`0` quirk and the "numbers outside 0–20 are rejected" claim match `parse_rating` (`initialise_dictionaries.py:65-86`), including the whitespace-stripping subtlety.
- The cache-marker contract (ignored on read → re-solved; marker written after a successful solve; "bump `CACHE_FORMAT_VERSION` when the value model changes") matches `read_write.py` and `initialise_dictionaries.py:27-56`.
- README's Input data section is accurate and pitched right for end users (the 15–5 example is a good touch).
- Stale-reference sweep: remaining mentions of `pairing_matrix.csv`/`map_importance` are only the intentional migration pointers (CLAUDE.md's "old-format folders" paragraph, the script itself) and PLAN.md's historical decision record. Nothing user-facing describes the old format as current.

## Scope

Clean: only CLAUDE.md, README.md, match-folder CSVs, and the new script — no engine or test files touched, and the golden suite passes unchanged. Archiving Scrim-Scotland-Old by deletion is reasonable: it's superseded by Scrim-Scotland, recoverable at `main` (and earlier tags), and keeping an unmigrated folder around would just be a loader trap.

Nice, tight PR — the full-cell verification came back spotless, and the PR description's claims all reproduced. Merge away.

svknoe:
Response to the review above:

**Finding 1 (score-0 round trip) — fixed** in e3c0812: `write_matrix` now emits `0.0` for an exact-zero score so it can never be re-read as the legacy even-matchup token. Unreachable from the folders migrated here (minimum converted score 3.5), but a correct catch for raw-numeric old matrices.

**Finding 2 (already-migrated guard skips cleanup)** — intentional: the guard makes re-runs a no-op instead of touching a folder that may have hand-edited new-format files. Anyone who wants the cleanup on such a folder can delete the best/worst CSVs and re-run.

**Finding 3 (asserts under -O)** — accepted as a known limitation of a one-shot migration script; the solver's own load path re-validates everything anyway (`validate_best_not_below_worst`, 0-20 range checks).

Thanks for the exhaustive 256-cell re-derivation. Merging with CI green.

PR #29, opened 2026-07-04
Honest CSV validation errors

### LINKED ISSUE #11: Honest CSV validation errors

svknoe:
Replaces the bare `except:` in `initialise_input_dictionary` that masked every failure as `SystemError: Missing file`. Closes #11.

## What changes

Every failure mode now raises `InputError` (a `ValueError` subclass) whose message names the actual problem and location:

| Failure | Example message |
|---|---|
| Missing file | `Missing input file: …\NoSuchTeam\pairing_matrix_best.csv. A match folder needs both pairing_matrix_best.csv and pairing_matrix_worst.csv.` |
| Wrong encoding | `…pairing_matrix_best.csv is not UTF-8 encoded (…). Re-save the file as UTF-8.` |
| Unknown / out-of-scale rating | `…, line 5, column 3 (Carol vs Chaos): unknown rating 'x'. Use a 0-20 score or one of the legacy tokens --/-/0/+/++.` |
| Ragged row | `…, line 4 (Bob): 3 ratings for 4 enemies -- the matrix must be square.` |
| Non-square (rows/headers) | `…: 5 rating rows for 4 friendly players -- the matrix must be square.` |
| Unsupported team size | `…: 5 players per team; the draft needs 4, 6 or 8.` (previously surfaced much later from the dictionary-name helper) |
| Duplicate across teams | `…: name(s) present on both teams: Alice. …` (still gated on `settings.require_unique_names`) |
| Duplicate within a team | `…: duplicate friendly player name(s): Alice. Names within a team must be unique.` (new — previously overwrote a row silently) |

Line numbers are the real CSV line numbers (blank lines are skipped without shifting them — the old loop's row index drifted after a blank line). Header names are stripped of surrounding whitespace. The `hard_crash` parameter (dead since both files became required in #27) and the module's last `loguru`/`sys.exit` usages are gone; `validate_best_not_below_worst` raises `InputError` too.

Note: `loguru` is now entirely unused in the codebase — I left the dependency pin in pyproject.toml alone since removing it is packaging, not validation; say the word if you'd like a follow-up.

## Tests

New `drafter/tests/test_input_validation.py` covers every failure mode above plus the happy path and blank-line/line-number behaviour (13 tests, in-process with a monkeypatched `get_path`). The mixed tokens/scores test grew to a legal 4-player matrix (the new player-count check rejects the old 2×2 fixture — a real bug in the test data, not a behavior regression).

## Acceptance

- [x] Each failure mode produces a message naming the actual problem and location (table above; all asserted in tests)
- [x] Smoke/golden tests unchanged: 48 passed, smoke draft complete

### DIFF
--- a/drafter/data/initialise_dictionaries.py
+++ b/drafter/data/initialise_dictionaries.py
@@ -1,7 +1,5 @@
 import time  # standard libraries
-import sys
 import csv
-from loguru import logger
 
 import drafter.common.utilities as utilities  # local source
 import drafter.common.team_permutation as team_permutation
@@ -13,8 +11,8 @@ import drafter.solver.game_state_dictionaries as game_state_dictionaries
 
 
 def initialise():
-    initialise_input_dictionary(match_info.pairing_dictionary_best, "pairing_matrix_best.csv", True)
-    initialise_input_dictionary(match_info.pairing_dictionary_worst, "pairing_matrix_worst.csv", True)
+    initialise_input_dictionary(match_info.pairing_dictionary_best, "pairing_matrix_best.csv")
+    initialise_input_dictionary(match_info.pairing_dictionary_worst, "pairing_matrix_worst.csv")
     validate_best_not_below_worst()
 
     if (settings.restrict_attackers):
@@ -56,6 +54,12 @@ def initialise():
         read_write.write_cache_format_marker(utilities.get_path(read_write.CACHE_FORMAT_FILENAME))
 
 
+class InputError(ValueError):
+    """A problem with a match folder's input CSVs, with a message that names
+    the actual file, location and cause (GitHub issue #11) -- as opposed to
+    the old behaviour of reporting every failure as 'Missing file'."""
+
+
 # Legacy rating tokens: expected 20-0 score margins. Kept so old-style matrices
 # and captains' shorthand keep working. Note that a bare '0' is a token (an
 # even matchup, i.e. an expected 10-10 score), not the 0-20 score 0.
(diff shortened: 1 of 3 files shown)


svknoe:
## Review

**Verdict: good to merge after one small fix** (a one-word change, below). The rewrite does what #11 asked: every failure mode from the issue — missing file, parse error with row/column, duplicate names across teams, non-square matrix, unknown rating token — now raises `InputError` with the actual file, location and cause. I read the full diff, checked all callers, probed edge cases empirically on the branch, and ran the fast suite: **48 passed** (plus the smoke fixtures load cleanly). The diff touches only `initialise_dictionaries.py` and tests — no solver code, so golden values are untouched by construction.

Things I specifically verified and found correct:

- **Line-number bookkeeping is right**, including the two hard cases: blank lines don't shift subsequent numbers (the old loop's index drift is really gone), and a quoted *multi-line* field in a header still yields correct physical line numbers for later data rows (probed: error correctly reported `line 5` when the first header spanned lines 1–2).
- **NUL bytes** (e.g. pasted binary junk) don't escape as a raw `_csv.Error` on Python 3.12 — they surface as a located `unknown rating '1\x001'` message. UTF-16 files (BOM `\xff\xfe`) hit the honest "not UTF-8 encoded" branch.
- **`hard_crash` removal is safe**: `initialise()` is the only production caller, both tests updated, nothing in `scripts/` uses it. `InputError(ValueError)` isn't swallowed anywhere — the `except ValueError` sites (`read_write.py` cache-marker read, `conftest.py` JSON scrape) are on unrelated paths.
- The expanded mixed tokens/scores matrix in `test_map_model.py`: I recomputed all 16 expected values by hand — all correct (including the `0` token vs `0.0` score distinction on Dave's row).
- Header stripping and the new duplicate/player-count checks don't affect any shipped match folder (all are clean 4/6/8-player BOM-less UTF-8 with unique, unpadded names).

### Must fix (one word)

**`path.open(encoding="UTF-8")` → `encoding="utf-8-sig"`** (initialise_dictionaries.py:118). Excel's "CSV UTF-8" — the most likely way a captain produces these files — writes a BOM, and `str.strip()` does **not** remove U+FEFF. Probed on the branch: a BOM'd but otherwise valid file loads with first ally `'﻿Alice'` — an invisible corruption that later surfaces as name-matching weirdness in the draft, or, if only one of the two files has a BOM, as a baffling "`pairing_matrix_worst.csv` is missing the Alice vs Ork entry" where both names *look* identical. Since this PR's whole purpose is honest input handling and the fix is one word (`utf-8-sig` decodes BOM-less UTF-8 identically and still raises `UnicodeDecodeError` for genuinely non-UTF-8 files — I checked), it belongs here. Worth a small test alongside it.

### Suggestions (nice-to-have, your call)

1. **Reject empty player names after stripping**, with a trailing-comma hint. Probed: an Excel-style trailing empty column on every line currently dies with `name(s) present on both teams: .` — an empty-string name rendered invisibly, which is the one genuinely misleading message left. A mid-row empty header cell (`Alice,,Carol,Dave`) is accepted silently as a player named `''`. A check like "empty player name in line 1, column 5 — check for a trailing comma" before the duplicate/overlap checks would cover both.
2. **`test_blank_lines_are_skipped_and_line_numbers_stay_real` doesn't test the second half of its name** — it only asserts the file loads. Add an error *below* the blank line and assert the reported line number is the physical one (that's the regression the old loop actually had).
3. **`raise InputError(...) from None`** in the two wrapping sites (UnicodeDecodeError, parse ValueError), so the CLI user sees one clean error instead of a chained "During handling of the above exception..." double traceback. Cosmetic but on-theme.

### Nits

- A *directory* named `pairing_matrix_best.csv` reports "Missing input file" (it isn't missing). `path.exists() and not path.is_file()` could special-case it; genuinely rare, fine to skip.
- The duplicate-within-team error can fire on a legitimate situation: two enemy players fielding the same faction, identified only by faction name in row 2. Rejecting is correct (dict keys can't represent it — the old code silently dropped a column), but a hint like "disambiguate, e.g. add the player's name" would save the one captain who hits it.
- Check ordering is otherwise sensible — for the common mistakes I tried (extra pasted column, missing header row, duplicated header row) the first error to fire names the real problem or points at the right line.

Not re-flagged per prior threads: the `discard_attacker` mixed-refusal bug (PR #27) and the now-unused `loguru` pin (deferred in the description — agreed it's a packaging follow-up).

svknoe:
Response to the review above — all addressed in b598d36:

- **Must-fix (BOM) — fixed:** reading with `utf-8-sig`, with a comment explaining the Excel "CSV UTF-8" trap; regression test added (BOM-prefixed file loads with a clean `Alice`).
- **Empty player names — added:** a trailing comma now reports `empty friendly player name in the header (check for a trailing comma or a missing name)` instead of the misleading both-teams overlap on `''`. Tested.
- **Blank-line test — strengthened:** it now actually asserts the reported line number (an error after a blank line points at the real file line).
- **`raise … from None` — done** for the two wrapped raises (unknown rating, encoding), so users get one clean error instead of a chained traceback.
- Directory-named-as-file reading as "Missing input file" and the duplicate-faction disambiguation hint: leaving as-is for now — both messages still name the right path/cause, and the second is a data-modelling choice more than an error-reporting one.

50 passed locally. Merging with CI green.

PR #31, opened 2026-07-04
Numeric ratings on the deviation scale (internal = score - 10)

### LINKED ISSUE #30: Fix numeric-rating conversion: internal scale is score-10, not 2*(score-10)
Correction to the rating-scale convention (PLAN.md §C, corrected 2026-07-04). The legacy tokens were always deviations from an even game: `+` = +4 ≡ a 14–6 game, `++` = +8 ≡ 18–2. The current `parse_rating` maps numeric 0–20 input as `2*(score-10)` while tokens keep face values — so the number `14` becomes +8 but the token `+` (the same 14–6 game) becomes +4. A matrix mixing digits and shorthand gets silently inconsistent values.

Changes:
- `parse_rating`: `score - 10` (drop the ×2). **Do not touch the token map** — it was always correct.
- **Do not touch the draft verdict thresholds** (±1/±4/±8 in draft.py) — they were designed on the deviation scale.
- Re-pin golden values. Verification fingerprints: a fixture whose CSVs are pure 0–20 numerics must come out *exactly halved*; a token-only fixture must be *unchanged*; a mixed fixture shifts non-uniformly (that is the bug being fixed — consider adding a small test asserting numeric `14` parses equal to token `+`).
- Docs: CLAUDE.md input-format section and any remaining `score = 10 + margin/2` phrasing → `score = internal + 10`. PLAN.md §C already corrected.
- Check `scripts/migrate_match_folder.py`: if it converts old-format numbers, ensure its output is consistent with the corrected conversion.
- This must land before M3/B1 re-pins goldens on the fast solver.

svknoe:
Fixes the numeric-rating conversion to the deviation scale per the corrected PLAN.md §C convention (owner-confirmed: `++` has always meant an 18–2 game). Closes #30.

## The fix

`parse_rating`: numeric 0–20 input now normalises as `internal = score − 10` (deviation from an even 10–10 game, range −10…+10). The token map is untouched — tokens were always deviations — so numeric `14` and token `+` (the same expected 14–6 game) finally parse identically. Draft verdict thresholds untouched (they were designed on this scale). `CACHE_FORMAT_VERSION` bumped to 3: caches solved under the 2×(score−10) reading are silently invalidated by the existing marker logic.

## Verification: the halving fingerprint

With the parse fix applied and the CSVs still in their old notation, a pure-numeric fixture's solve must come out exactly half — all payoff matrices scale by 0.5, which preserves zero-sum equilibria, and power-of-two scaling commutes with FP rounding:

| Fixture (old CSVs, new parse) | Old pin | Waypoint result |
|---|---|---|
| Smoke k=4 | 4.997553182223153 | 2.4987765911115765 — **bit-exact half** |
| Six k=4 | 1.5140341559225499 | 0.7570170779612748 — half to **1 ulp** |
| Scotland k=3 | 6.348743764113647 | 3.174371882056823 — half to **1 ulp** |

(The 1-ulp deviations on Six/Scotland show one nashpy code path isn't perfectly scale-equivariant; it's FP noise, not a logic difference.)

## Data re-migration (option 2, as agreed)

The #28 migration wrote `score = 10 + (pairing ± importance)/2` — correct under the margin reading, understating the intended spread by ×2 under the corrected semantics. All seven folders were re-generated from their pre-migration originals in git history with the corrected `scripts/migrate_match_folder.py`:

```
score = clamp(10 + pairing ± importance, 0, 20)
```

Clamping saturates at the 20–0 blowout (old `pairing ± importance` reaches ±13): Scotland 1 cell, Germany 2, Spain 4, Scrim-Scotland 3; Smoke/Six/Test none.

## Golden values

**Smoke and Six are deliberately bit-identical to the existing pins** — a strong consistency check, not an accident: the parse halving and the notation doubling cancel exactly wherever nothing clamps, proving both pipelines encode `pairing ± importance` internally. Re-verified with 3× bit-identical fresh solves each, plus the independent scipy-linprog brute force on Smoke (updated to the corrected conversion): `4.997553182223154`, same 1-ulp agreement and identical strategy as in #27.

**Scotland is bit-identical too** (2× fresh solves: `6.348743764113647`, same pure Petter/Drukhari strategies): its one clamped cell (Bjørn vs Sisters worst map, deviation −13 → −10) never reaches the k=3 equilibrium path. Net result: **no golden value changes anywhere** — the strongest possible evidence that this PR is a pure notation/semantics fix.

## Also in this PR

- CLAUDE.md input-format section: deviation-scale wording (`internal = score − 10`, token examples `+` ≡ 14–6, `++` ≡ 18–2)
- New unit test pinning `parse("14") == parse("+")` (the exact bug class this issue fixes), all parse tests moved to the deviation scale
- Migration-script docstring documents the corrected formula and the clamp rationale

## Acceptance

- [x] Numeric and token input agree (unit-tested)
- [x] Halving fingerprint verified on all three fixtures before re-migration
- [x] Golden values confirmed unchanged with provenance (fast suite: 51 passed; `pytest -m slow`: 1 passed, 244s)
- [x] Lands before M3/B1 re-pins goldens on the fast solver


svknoe:
## Review: deviation-scale correction (#30)

**Verdict: good to merge.** The fix is exactly what issue #30 asked for, the re-migrated data is bit-exact against the pre-migration originals, the unchanged golden pins are the *right* outcome (and well argued), and I independently re-verified the key claims on this machine. No must-fix findings; two cosmetic nits below.

### What I verified independently

- **`parse_rating`** (`drafter/data/initialise_dictionaries.py`): `score - 10` with the token map and the 0–20 range check untouched — correct and complete. I grepped the whole repo for leftover `2*(score-10)` logic and margin-scale assumptions: the only remaining mentions are deliberate historical notes (cache-version comment, golden-value provenance comments, PLAN.md's "earlier margin-based reading" parenthetical). No live code path still doubles. README's input-format text is scale-agnostic and stays correct.
- **Re-migration, all seven folders**: I scripted the check rather than spot-checking — parsed each folder's `pairing_matrix.csv` + `map_importance_matrix.csv` at `b58512b`, applied `clamp(10 + pairing ± importance, 0, 20)`, and diffed against the CSVs on this branch. **Every cell of every best/worst matrix matches exactly**, and the clamp counts are precisely as claimed: Scotland 1, Germany 2, Spain 4, Scrim-Scotland 3, Smoke/Six/Test 0. Manual spot checks agree, including the ones worth eyeballing: Smoke `++`/imp 1 → 19/17, Scotland Bjørn vs Sisters (`--`, imp 5) → best 7, worst −3 → **0.0** as claimed, Test `++` → 18 with best ≡ worst.
- **Fast suite**: `51 passed, 1 deselected in 13.15s` on this branch.
- **Slow suite**: `pytest -m slow` re-run here — passes, i.e. the Scotland pin `6.348743764113647` reproduces on the re-migrated data with the corrected parse. That's the one pin a clamp could have touched, so this is now verified on two machines/runs beyond the PR's own 2×.
- **Test arithmetic**: every updated expectation checks out by hand — `test_input_validation` Dave-vs-Ork 7 → −3; the mixed-fixture dictionary (`++,15,0,8.5` → 8, 5.0, 0, −1.5 etc.); the new `parse("14") == parse("+") == 4` / `parse("2") == parse("--") == −8` test is exactly the regression guard issue #30 asked for. `0.0` → −10 and the whitespace-stripping cases are all consistent with the new scale.

### On the unchanged golden values

The reasoning is sound, and it's stronger than "the numbers happened to agree." Old pipeline: files stored `10 + dev/2`, parse doubled → internal `dev`. New pipeline: files store `10 + dev`, parse subtracts 10 → internal `dev`. Halves and small integers are exact in binary floating point, so wherever nothing clamps, the payoff matrices are **bit-identical**, and identical inputs to a deterministic solver must give bit-identical outputs — the Smoke/Six pins *had* to survive, and their survival is a genuine end-to-end consistency proof of both the old and new encodings. The halving fingerprint (run against the old CSVs with the new parse before re-migrating) is a nice independent cross-check, and the 1-ulp wobble on Six/Scotland is the expected signature of a nashpy path that isn't perfectly scale-equivariant — a ×0.5 that survives bit-exactly on one fixture and to 1 ulp on the others is FP noise, not a semantics difference. Scotland's clamped cell (−13 → −10) was the only way a pin could move; the claim that it doesn't reach the k=3 equilibrium is verified empirically by the bit-identical re-solves (PR's 2× + my run above) rather than argued a priori, which is the right standard here.

One observation for the record: the old data encoded a deviation of −13, which is unrepresentable under the corrected semantics (you can't lose worse than 0–20) — so the clamp isn't lossy compression, it's removing a value that was never physically meaningful. Worth remembering if a future k > 3 solve on Scotland shifts slightly: that would be the clamp becoming visible, not a regression.

### Cache invalidation

`CACHE_FORMAT_VERSION = 3` in `drafter/data/read_write.py` is correct and sufficient: `cache_format_is_current` does an exact equality check on the marker, so any folder stamped 2 (or unstamped) falls back to a fresh solve with a printed notice, and the marker is only re-written after a successful solve+write. No cache JSONs are committed to the repo, so this only affects local user caches — exactly what the mechanism is for. The version-history comment documenting v2 as "the mistaken 2*(score-10) conversion" is good archaeology.

### Docs

CLAUDE.md's rewritten input-format paragraph matches the code exactly (`internal = score − 10`, range −10…+10, token examples ≡ 14–6 / 18–2, bare-`0` quirk retained). The migration-script docstring now states the corrected formula and the clamp rationale, and the runtime assert-on-out-of-scale was correctly replaced with clamp-and-report. Verdict thresholds in `drafter/solver/draft.py` (±1/±4/±8) untouched, as instructed — and they genuinely read correctly on the deviation scale, since a difference of achieved-vs-expected deviation sums *is* a difference in 20-0 game points.

### Nits (take or leave)

1. `drafter/tests/test_map_model.py:2` — the module docstring still says "rating normalisation to the internal **margin** scale"; everything else in the file was moved to deviation wording.
2. `drafter/tests/test_map_model.py:22-24` — `test_parse_rating_legacy_tokens` still parametrizes `token,margin`; the values are unchanged (tokens were always right), but the name is now the wrong word.

Neither blocks merge; both could ride along with any future commit touching that file.

svknoe:
Response to the review above:

Both nits fixed in 4f235a1 (margin → deviation wording in the test-module docstring and the token parametrize name). Thanks for the independent full re-derivation of all seven folders against the b58512b originals and for reproducing the Scotland pin a fourth time — with the reasoning verified that the unchanged pins were mathematically forced for unclamped cells, this correction lands with the golden baseline provably intact.

Merging with CI green.

Issue #12, opened 2026-07-03
B1: solve zero-sum games directly (closed form + LP)


svknoe:
Every game in the tree is zero-sum, but is solved with nashpy support enumeration and its degenerate-game fallback chain (support → vertex → Lemke-Howson). Replace with: 2×2 closed form (~850k of the ~950k solves at k=3), larger matrices via one scipy linprog (HiGHS) call. Keep the matrix-hash cache. Remove the nashpy dependency.

Acceptance: golden values unchanged (strategy mixes may differ between equivalent equilibria — values must match); Scotland k=3 strategy phase < 30 s (baseline 275 s); no RuntimeWarnings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #33, opened 2026-07-04
discard_attacker: return refused attackers to the child pools

### LINKED ISSUE #32: discard_attacker (6/8-player) wires the kept attackers into the child game instead of the refused ones
Pre-existing solver bug on main, found during the PR #27 review and independently confirmed twice (reviewer's brute force, implementer's re-derivation) plus a third manual re-derivation. Full analysis in the PR #27 comment thread.

In `drafter/solver/games.py::discard_attacker` (n=6/8 path only — the n=4 closed form and the diagonal entries are correct), the mixed-refusal payoff entries pass the *played* attackers back to the child-game pools instead of the *refused* ones:

- `AB` uses `get_game_key(f_attacker_A, e_attacker_B)` but should use `get_game_key(f_attacker_B, e_attacker_A)`
- `BA` is the mirror image

The kept attackers are double-counted (they play the defender now *and* continue in the child game) while the refused pair never plays again. Measured effect (Six fixture, unrestricted): engine `1.5180657…` vs rules-correct `1.9058104…` — ~0.39 internal points at the top level, and slightly wrong continuations in every real 6/8-player solve.

Fix is a two-line swap, plus:
- re-pin Six and Scotland goldens (Smoke is 4-player: provably unaffected)
- bump `CACHE_FORMAT_VERSION` (existing caches bake the bug in)
- **must land before B1 (#12)** so the LP rewrite's 'golden values unchanged' criterion verifies against rules-correct values — do this as M3's first PR

The reviewer offered their brute-force script (explicit recursion + scipy linprog) in the PR #27 thread — worth committing alongside as an independent oracle; it reproduces the Smoke golden to 1 ulp.

svknoe:
Fixes the pre-existing discard-stage bug found in the PR #27 review: the 6/8-player `discard_attacker` wired the *kept* attackers into the child gamestate pools instead of the *refused* ones, double-counting the players who just played and never letting the refused pair play again. Closes #32. M3's first PR, landing before B1 (#12) so the LP rewrite verifies against rules-correct values.

## The fix

Two child keys swap in `games.py::discard_attacker` (n=6/8 path): `AB` now returns `(f_attacker_B, e_attacker_A)` to the pools, `BA` the mirror. The diagonal entries and the 4-player closed form were always correct. A comment now states the invariant. `CACHE_FORMAT_VERSION` → 4: every existing cache baked the bug in and is silently invalidated by the marker logic.

## Verification

**Committed oracle.** `scripts/brute_force_oracle.py` — an independent implementation (explicit 4/6-player draft recursion with memoised subgames, zero-sum games via `scipy.optimize.linprog`, no drafter solver code) grown from the #27/#30 verification scripts. 8-player folders are rejected as impractical; the 6-player tree exercises the recursive discard wiring that a 4-player oracle cannot see (which is why #27's Smoke-only check missed this bug).

| Check | Oracle | Fixed engine |
|---|---|---|
| Six, unrestricted | `1.9058104884743514` | `1.9058104884743512` (1 ulp, same mix Cleo 0.103 / Elin 0.897) |
| Smoke (4-player, fix can't apply) | `4.997553182223154` | `4.997553182223153` (existing pin, 1 ulp) |

The oracle's unrestricted-Six value also reproduces the #27 reviewer's rules-correct number (`1.9058104884743514`) to the digit, from a fully separate implementation.

**Wiring regression test.** New `drafter/tests/test_discard_wiring.py` pins the child-key wiring itself: sentinel child values per (returned friend, returned enemy) pair, payoff matrix captured via a stubbed `get_game_strategy`, every cell asserted. A re-swap would fail this test directly rather than surfacing as an unexplained golden shift.

## Golden re-pins (deliberate — this is the point of the fix)

| Fixture | Old pin (encoded the bug) | New pin | Provenance |
|---|---|---|---|
| Smoke k=4 | `4.997553182223153` | **unchanged** (provably: 4-player) | 3× bit-identical |
| Six k=4 | `1.5140341559225499` | `1.9058104884743512` | 3× bit-identical + oracle to 1 ulp; equals the unrestricted value (the k-heuristic doesn't bind on this fixture) |
| Scotland k=3 | `6.348743764113647` | `5.875946490218083` | 2× bit-identical fresh solves + `pytest -m slow` (a third) |

**The Scotland equilibrium changed, not just the value**: the friendly defender flips from pure Petter to pure **Mariusz** (enemy stays pure Drukhari), and the value drops by ~0.47 — the bug's double-counted kept attackers were inflating continuations enough to recommend the wrong defender on a real matrix.

Fast suite: 52 passed. Smoke draft complete. `pytest -m slow`: 1 passed, 235s.

### DIFF
--- a/drafter/data/read_write.py
+++ b/drafter/data/read_write.py
@@ -4,11 +4,12 @@ import json  # standard libraries
 # Caches carry solved game values, so any change to the value model makes old
 # caches silently wrong (they would load fine and give plausible numbers).
 # Bump this whenever cached values are no longer comparable with the current
-# engine. Version 3 = numeric ratings on the deviation scale, internal =
-# score - 10 (issue #30); version 2 = the 11th-edition best/worst map model
-# (issue #9) with the mistaken 2*(score-10) conversion; version 1 (implicit --
-# no marker file existed) = the old map-importance model.
-CACHE_FORMAT_VERSION = 3
+# engine. Version 4 = discard_attacker returns the refused attackers to the
+# child pools (issue #32); version 3 = numeric ratings on the deviation scale,
+# internal = score - 10 (issue #30); version 2 = the 11th-edition best/worst
+# map model (issue #9) with the mistaken 2*(score-10) conversion; version 1
+# (implicit -- no marker file existed) = the old map-importance model.
+CACHE_FORMAT_VERSION = 4
 CACHE_FORMAT_FILENAME = "cache_format.json"
 
 
--- a/drafter/solver/games.py
+++ b/drafter/solver/games.py
@@ -112,9 +112,13 @@ def discard_attacker(n, selected_attackers_gamestate, select_defender_strategies
     fA_eD = utilities.get_pairing_value(f_attacker_A, e_defender, e_defender)
     fB_eD = utilities.get_pairing_value(f_attacker_B, e_defender, e_defender)
 
+    # The child gamestate receives the REFUSED attackers back into the pools
+    # (issue #32): in AB, e_B plays the friendly defender and f_A plays the
+    # enemy defender, so the refused pair returning to the pools is (f_B, e_A)
+    # -- and mirrored in BA. The kept attackers' games are the fD_*/f*_eD terms.
     AA = fD_eB + fB_eD + select_defender_strategies[get_game_key(f_attacker_A, e_attacker_A)][2]
-    AB = fD_eB + fA_eD + select_defender_strategies[get_game_key(f_attacker_A, e_attacker_B)][2]
-    BA = fD_eA + fB_eD + select_defender_strategies[get_game_key(f_attacker_B, e_attacker_A)][2]
+    AB = fD_eB + fA_eD + select_defender_strategies[get_game_key(f_attacker_B, e_attacker_A)][2]
+    BA = fD_eA + fB_eD + select_defender_strategies[get_game_key(f_attacker_A, e_attacker_B)][2]
     BB = fD_eA + fA_eD + select_defender_strategies[get_game_key(f_attacker_B, e_attacker_B)][2]
 
     game_array = np.array([[AA, AB], [BA, BB]])
--- a/drafter/tests/test_discard_wiring.py
+++ b/drafter/tests/test_discard_wiring.py
@@ -1,0 +1,86 @@
+"""Regression test for the discard_attacker child-key wiring (issue #32).
+
+The n=6/8 discard game's off-diagonal entries must return the REFUSED
+attackers to the child pools, not the attackers who just played. The golden
+tests would catch a regression only as an unexplained value shift; this test
+pins the wiring itself, cell by cell, with sentinel child values.
+
+In-process like test_map_model.py: match_info is monkeypatched and
+get_game_strategy is stubbed to capture the payoff matrix, so no solver or
+global state is involved.
+"""
+import pytest
+
+import drafter.common.utilities as utilities
+import drafter.data.match_info as match_info
+import drafter.solver.games as games
+from drafter.common.draft_stage import DraftStage
+from drafter.common.game_state import GameState
+from drafter.common.team_permutation import TeamPermutation
+
+
+def child_key(remaining_friends, extra_friend, remaining_enemies, extra_enemy):
+    return GameState(
+        DraftStage.none,
+        TeamPermutation(remaining_friends + [extra_friend]),
+        TeamPermutation(remaining_enemies + [extra_enemy]),
+    ).get_key()
+
+
+def test_discard_attacker_returns_refused_attackers_to_child_pools(monkeypatch):
+    # 6-player state, mid-draft: defenders picked, attackers presented.
+    # Names chosen already-sorted so TeamPermutation's attacker sorting is a
+    # no-op and fA/eA keep their roles.
+    f_defender, f_attacker_A, f_attacker_B = "Fdef", "FattA", "FattB"
+    remaining_friends = ["Frem1", "Frem2", "Frem3"]
+    e_defender, e_attacker_A, e_attacker_B = "Edef", "EattA", "EattB"
+    remaining_enemies = ["Erem1", "Erem2", "Erem3"]
+
+    # Distinct pairing values so each fixed-pairing term is identifiable:
+    # the friendly defender defends (best map), the enemy defender forces the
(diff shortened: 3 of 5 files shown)


svknoe:
# Review: discard_attacker child-key fix (#32)

**Verdict: good to merge.** I re-derived the payoff semantics from scratch, re-ran both oracle checks and the fast suite locally, and confirmed the new wiring test actually discriminates by reverting the fix and watching it fail. No must-fix items. Two take-or-leave nits on the oracle script at the bottom.

## 1. The fix — independently re-derived, correct

I rebuilt the 2×2 from first principles, using only `get_game_strategy`'s option labels (`rows = [e_attacker_A, e_attacker_B]` = which enemy attacker the friendly side refuses; columns mirror) and the pairing terms:

- **AB** (friendly refuses e_A, enemy refuses f_B): kept attackers are e_B (plays f_defender → `fD_eB`) and f_A (plays e_defender → `fA_eD`). The pairing terms already consume the kept pair, so the child pools must receive the refused pair **(f_B, e_A)** → `get_game_key(f_attacker_B, e_attacker_A)`. Matches the fixed line.
- **BA** is the exact mirror: kept (e_A, f_B), refused **(f_A, e_B)**. Matches.
- **AA/BB**: kept and refused coincide per side symmetrically; the original keys were already the refused pair. Correctly untouched.
- **`discard_attacker_4`**: the closed form pairs refused-vs-refused (`fB_eA` in AB etc.) consistently with the same row/column reading. Correctly untouched.

The reading is corroborated by the rest of the codebase: `get_none_team_permutation` returns `remaining + [discarded_attacker]` (the refused player rejoins the pool), and `draft.py`'s discard step applies the *opponent's* selection as each team's `discarded_attacker` before calling it — so the enumeration path always had the rules right, and `discard_attacker`'s hand-built child keys now agree with it. Both off-diagonal child keys exist under enumeration (the discard cartesian product generates all four `(extra_friend, extra_enemy)` combinations, k-restricted or not), which is also why the bug was silent rather than a KeyError. The new invariant comment is accurate.

## 2. Oracle — faithful to the rules; claimed values reproduced on this machine

Read end to end: defender stage (simultaneous, friendly defender → best map, enemy → worst, neutral midpoint hard-coded at 0.5 matching `settings.neutral_map_weight`) → attackers stage over C(rest, 2) per side → discard 2×2 with the **refused** attackers rejoining the pools in the recursive branch, and the closed refused-vs-refused + last-vs-last endgame in the 4-player branch. Stage order and simultaneity match the engine; the LP formulation (positivity shift + `min Σx, Aᵀx ≥ 1`) is the standard zero-sum construction and returns the exact game value. Memoisation is sound: `none_stage_value` is keyed on the two sorted player pools, which — given the fixed global matrices — fully determine a fresh round's value. The 8-player rejection is explicit and documented.

Ran both:

| Fixture | My run | PR claim |
|---|---|---|
| Six (unrestricted) | `1.9058104884743514`, mix Cleo 0.103 / Elin 0.897 | same, to the digit |
| Smoke | `4.997553182223154` | same; 1 ulp from the engine pin `…153` |

## 3. Wiring test — verified it discriminates, not just passes

I locally reverted the two `games.py` lines to the old wiring and ran `test_discard_wiring.py`: it fails with exactly the child-sentinel swap visible (`[1022.0, 3012.0] != [1022.0, 2012.0]` — the 2000/3000 sentinels trading places), then restored the fix and it passes. The stubbing is correctly wired (`games.py` reads `utilities.get_game_strategy` and `match_info.pairing_dictionary_*` through their modules at call time, so `monkeypatch.setattr` on those modules is seen), the −99 sentinels would catch a defender-dispatch mistake, the 1000-spaced child sentinels can't collide with the small pairing terms, and the pre-sorted names neutralise `TeamPermutation`'s attacker sorting. Good test — it pins the invariant, not a number.

## 4. Golden re-pins

- **Fast suite: 52 passed, 1 deselected, ~13 s** locally; CI green on the branch.
- **Six** `1.9058104884743512` at k=4: the passing golden plus my unrestricted oracle run at `…514` directly evidences the "k-heuristic doesn't bind" claim — the k=4 engine value equals the unrestricted rules-correct value to 1 ulp with the same top-level mix. (k=4 does *restrict the option set* at the 6-player attackers stage, 10 combos → 6; it just excludes nothing in equilibrium support on this fixture. The claim is stated correctly as fixture-specific.)
- **Smoke** unchanged — provable, as stated: the n=4 path never executes the fixed code, and my oracle run agrees with the pin to 1 ulp.
- **Scotland** `5.875946490218083` / Petter → Mariusz: not re-run here (per the documented 3× provenance — two bit-identical fresh solves plus the slow test). The direction is plausible: the bug double-counted kept attackers into continuations, inflating exactly the kind of defender whose refused continuations were being overvalued, and a ~0.47 top-level shift is in line with the ~0.39 measured on Six. The updated provenance comments in `test_golden_values.py` are accurate about what changed and why.

## 5. Cache version bump — correct and sufficient

`cache_format_is_current` requires exact equality with `CACHE_FORMAT_VERSION`, so every existing marker (v3 or older/absent) now fails the check and the cached gamestate/strategy JSONs are ignored and re-solved. No `cache_format.json` is committed anywhere (verified `git ls-files`), so nothing in-repo needs touching, and the in-memory nashpy caches are process-local. The version-history comment is a nice touch.

## Nits (take or leave)

- `scripts/brute_force_oracle.py` calls `main()` unguarded at module bottom. An `if __name__ == "__main__":` guard is one line and prevents surprise execution if the file is ever imported (e.g. by accidental test collection). Cosmetic — it lives in `scripts/`.
- The oracle's `read_matrix` doesn't assert `best >= worst` (the engine validates this on load). A one-line assert would stop a mis-edited fixture from producing an oracle value the engine could never agree with. Optional hardening for a verification script.

Nothing else. The fix is the minimal two-line swap the issue called for, the verification is genuinely independent and reproduces on this machine, and landing it before B1 (#12) is the right sequencing.

svknoe:
Response to the review above:

Both nits taken in 67a6a39: the oracle now has an `if __name__ == "__main__":` guard and asserts `best >= worst` right after reading (re-verified against Smoke afterwards). Particularly appreciated: locally reverting the two fixed lines to prove the wiring test discriminates (sentinels visibly swapped), and reproducing both oracle values from scratch — with your independent re-derivation this key swap has now been confirmed by four separate derivations across three sessions.

Merging with CI green. B1 (#12) can now verify against rules-correct goldens.

PR #34, opened 2026-07-05
B1: solve zero-sum games directly (closed form + LP)

### LINKED ISSUE #12: B1: solve zero-sum games directly (closed form + LP)

svknoe:
Closes #12.

Every game in the draft tree is zero-sum, but was solved with nashpy's general-bimatrix support enumeration and its degenerate-game fallback chain (support → vertex → Lemke-Howson) — slow on the discrete rating scale, and it prints "degenerate game" RuntimeWarnings. This replaces it with a direct zero-sum solver in `utilities.get_game_solution`.

## What changed

`get_game_solution(game_array)` now solves a zero-sum matrix game (row = friendly/maximiser, column = enemy/minimiser) directly:

1. **Pure saddle point** when one exists — `maximin == minimax` over pure strategies. On the Scotland k=3 tree ~half of the non-2×2 games have one, so no LP is needed for them.
2. **2×2 closed form** otherwise (the ~90% common case). Entries are pulled out with `.tolist()` so it's plain-float arithmetic.
3. **One `scipy.optimize.linprog` (HiGHS) call** otherwise. The column player's strategy is read from the LP **dual** (the inequality-constraint marginals), so a single solve yields both sides — no symmetric second LP.

Two matrix-hash caches:
- the existing per-stage bit-identical caches in `games.py`;
- a new **translation-normalised** cache (`normalised_game_solution_cache`) — zero-sum strategies and value are invariant under adding a constant to every payoff, so matrices that differ only by a constant (common on the discrete scale) share one LP solve (~21% fewer LP solves on Scotland k=3).

Removes the `nashpy` dependency, and `networkx` (only nashpy pulled it in — nothing imports it).

## Acceptance criteria (issue #12), on the reference machine (Ryzen 9800X3D)

| Criterion | Result |
|---|---|
| Golden values unchanged | ✅ Scotland k=3 top-level value **bit-identical** (`5.875946490218083`), support sets unchanged; Smoke/Six exact; slow test green |
| Scotland k=3 strategy phase < 30 s (baseline 275 s) | ✅ **~28.5–29.2 s** (≈9×) |
| No RuntimeWarnings | ✅ stderr clean |
| Remove nashpy | ✅ removed from `pyproject.toml`; no code imports it |

## Tests

- New `drafter/tests/test_zero_sum_solver.py`: closed-form / saddle / LP paths, each checked by the **equilibrium security property** (`min_j x·A[:,j] ≥ v ≥ max_i A[i,:]·y`) rather than pinned strategy vectors — so it's equilibrium-selection agnostic while pinning the (unique) value.
- Full fast suite: **70 passed**. Slow Scotland golden test: **passed**.

## Notes

- The remaining strategy-phase cost is split between the ~35k genuinely-mixed LP solves (scipy's `linprog` per-call *wrapper* overhead, ~250 µs, dominates for these tiny 3×3 games) and the string-keyed gamestate iteration — the latter is what **B2 (integer state encoding)** targets. A pure-saddle fast path was added (beyond the issue's literal "2×2 closed form + LP") to halve the LP count and clear the 30 s target.
- Cache format marker deliberately **not** bumped: the value model is unchanged (equilibrium values are unique), so caches solved under nashpy load with correct values.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,7 +32,7 @@ drafter/
     team_permutation.py    one team's partial draft state; enumeration of successor states;
                            the k-restriction heuristic (restrict to k best attackers)
     game_state.py          GameState = (draft_stage, friendly TeamPermutation, enemy TeamPermutation)
-    utilities.py           nashpy game solving, pairing-value lookup, key/dictionary naming
+    utilities.py           direct zero-sum game solving (2x2 closed form + LP), pairing-value lookup, key/dictionary naming
   data/
     settings.py            all knobs (team name, k-restriction, read/write caches)
     match_info.py          module-level globals: enemy team name + input matrices
@@ -61,9 +61,15 @@ drafter/
 2. **Strategy solving** (`strategy_dictionaries`): iterate stages deepest
    first (4-player select_attackers up to 8-player none). For each gamestate
    build the payoff matrix whose entries are the already-solved values of
-   child states (`games.py`), then solve that matrix game with **nashpy**
-   (support enumeration → vertex enumeration → Lemke-Howson fallbacks, cached
-   by matrix hash). The 4-player discard endgame is closed-form.
+   child states (`games.py`), then solve that matrix game **directly as the
+   zero-sum game it is** (`utilities.get_game_solution`, PLAN.md B1): a pure
+   saddle point when one exists, otherwise a 2x2 closed form, otherwise one
+   `scipy.optimize.linprog` (HiGHS) call whose dual yields the opposing
+   strategy. Deterministic, no degenerate-game warnings. Solutions are cached
+   by matrix hash — both bit-identical (per-stage caches in `games.py`) and
+   translation-normalised (`utilities.normalised_game_solution_cache`, since
+   zero-sum strategies/value are invariant under adding a constant). The
+   4-player discard endgame is closed-form.
 3. **Draft loop** (`draft.py`): walks the solved tree interactively. If the
    user makes a move outside the enumerated tree (possible with k-restriction),
    the tree is extended and re-solved on the fly.
@@ -76,12 +82,16 @@ and last-players games).
 
 **Known scale/pain points** (measured 2026-07, Ryzen 9800X3D, k as noted):
 
-- Fresh 8-player solve is the "one hour, 16 GB RAM" problem the user reports
-  (historical, k=4). The cost is dominated by the ~10^5–10^6 4-player-stage
-  gamestates, each solved by nashpy support enumeration; the discrete rating
-  scale makes most games degenerate, triggering slow fallbacks.
-- All games here are **zero-sum**, so each could be solved by one small LP
-  instead of enumeration — the main known speed lever (see PLAN.md).
+- Fresh 8-player solve was historically the "one hour, 16 GB RAM" problem
+  (k=4). The cost is dominated by the ~10^5–10^6 4-player-stage gamestates.
+- **B1 (done, issue #12):** every game is zero-sum, now solved directly
+  (saddle / 2x2 closed form / one HiGHS LP) instead of by nashpy's
+  general-bimatrix support enumeration and its degenerate-game fallback chain.
+  This cut the Scotland k=3 strategy phase from ~275 s to ~29 s (≈9×) with
+  bit-stable golden values and no warnings. The remaining strategy-phase cost
+  is now split between the ~35k genuinely-mixed LP solves (scipy's per-call
+  wrapper overhead dominates for these tiny games) and the string-keyed
+  gamestate iteration — the latter is what B2 (integer state encoding) targets.
 - Cached JSON for one 8-player opponent ≈ 630 MB (strategy dicts dominate);
   loading that cache takes ~19 s. String keys are a large share of the RAM.
 - `restricted_attackers_count` (k) in settings.py is the only current knob:
@@ -154,8 +164,10 @@ Notes for agents:
   same explicit call.
 - Settings are plain module globals in `drafter/data/settings.py`; there are
   no CLI flags. Monkeypatch settings before `initialise()` in scripts.
-- nashpy prints "degenerate game" RuntimeWarnings during solving — harmless,
-  the code falls back to other algorithms.
+- The zero-sum solver is deterministic and silent: no RuntimeWarnings during
+  solving (nashpy's "degenerate game" warnings are gone with nashpy, issue
+  #12). scipy's `linprog` is a required dependency; nashpy and networkx are
+  no longer dependencies.
 
 ## Conventions & state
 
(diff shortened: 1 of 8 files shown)


svknoe:
_Automated one-pass code review (Claude Code, Opus 4.8). Ran the fast suite + solver unit tests + a 20k-case randomized solver fuzz + an exhaustive 2×2 grid check locally; did **not** run the slow Scotland solve._

## Verdict: Approve

Solid, correct, well-documented change. The direct zero-sum solver is mathematically sound on every path I could exercise, the interface contract is preserved, and the nashpy/networkx removal is clean. Findings below are all non-blocking; the two "should-fix" items are latent footguns, not present-day bugs.

I independently verified:
- **Dual-marginals column strategy is reliable**, including degenerate cases. 20,000 randomized discrete-rating games (2–5 dims, integer payoffs — the regime that made nashpy degenerate) and an exhaustive 2×2 grid: **0 failures** of the equilibrium security property, non-negativity, or normalization.
- **Saddle detection via exact `==` is sound and conservative.** Over ~106k detected-saddle cases, `a[best_row, best_column]` always equals the common maximin/minimax value (as it must: `a[br,bc] ≥ rmin[br] = V` and `≤ cmax[bc] = V`). Float noise can only make a genuine saddle *miss* the `==` and fall through to the LP — which solves it correctly regardless. It cannot misclassify a non-saddle as a saddle.
- **2×2 closed form** (incl. the saddle fallback when the mixed formula leaves [0,1] or denominator == 0) is valid across the full grid.
- **Interface/behavior preserved**: `get_game_solution` and `get_game_strategy` return shapes unchanged; the `test_discard_wiring.py` stub signature still matches; child consumers read only `[2]` (value). Rounding moved to cache-store time is bit-identical output (`round(.,3)` idempotent; value never rounded, then or now).
- **Acceptance criteria**: golden values bit-identical (Smoke/Six pass locally; Scotland pinned in the slow test); nashpy removed and no imports remain; networkx only referenced inside the (now-removed) nashpy package, so its removal is safe; `scipy==1.14.0` already declared, so the new hard import is covered; solver is deterministic/silent (no RuntimeWarnings).

## Should-fix

- **`normalised_game_solution_cache` key is shape-blind** (`utilities.py:135`). `positive_a.tobytes()` does not encode shape, so a post-shift `2×N` and `N×2` (or any two same-total-size matrices with the same C-order element sequence) collide to one key and return a solution of the wrong shape. This cache is **global across all stages and sizes**, unlike the per-stage `games.py` caches whose shapes are homogeneous — so it's the more exposed one. `get_game_strategy`'s length guard would catch a *friendly-dimension* mismatch but not a matching-count transpose. Extremely unlikely in the real tree, but cheap to close: fold `positive_a.shape` (and ideally `dtype`) into the key, e.g. `hash((positive_a.shape, positive_a.tobytes()))`.

- **Cached strategy lists are aliased to callers** (`utilities.py:141–143`). Two translation-equivalent matrices receive the *same* list objects, which also alias the cache entry (confirmed: `sol_a[0] is cached[0]`). Safe today — every consumer treats them as read-only and `get_game_strategy` rebuilds rounded lists via comprehension — but it's a footgun: any future in-place mutation of a returned row/column strategy would silently corrupt other gamestates' cached solutions. Consider returning copies, or a one-line comment stating the invariant that these lists must never be mutated in place.

## Nits

- **`.tostring()` is deprecated** (`get_game_strategy`, `utilities.py:170`). Pre-existing (the PR only renamed variables around it), but the PR newly introduces `.tobytes()` at line 135, so the two hash sites are now inconsistent. `.tostring()` emits a `DeprecationWarning` (suppressed by default, hence "stderr clean" holds, but visible under `-W`) and is gone in later numpy. Swap to `.tobytes()` for consistency and future-proofing.

- **No stored-matrix re-check guards a genuine 64-bit `hash()` collision** in either cache (`utilities.py:135–139`, and the pre-existing `games.py` sites). Astronomically unlikely; noting for completeness only.

- **Acceptance margin is thin.** PR reports ~28.5–29.2 s against the 30 s bar on the reference machine — comfortably passing but with little headroom on slower hardware. Not a code issue; just flagging that the "< 30 s" criterion is met by ~1 s.

## Notes on PR-body claims

The distributional claims ("~90% are 2×2", "~half of non-2×2 have a saddle", "~21% fewer LP solves") are performance characterizations I couldn't independently verify without the full solve, but they're plausible and not correctness assertions. "One solve yields both sides via the dual — no symmetric second LP" is architecturally correct and matches my dual-marginals verification. "Cache format marker deliberately not bumped" is justified: the value model is unchanged and equilibrium values are unique, so nashpy-solved caches load with correct values.

svknoe:
Thanks for the thorough pass (incl. the 20k fuzz + exhaustive 2×2 grid). Dispositions, all addressed in ce81e90:

**Should-fix — shape-blind cache key** — Fixed. The normalised cache now keys on `(positive_a.shape, positive_a.tobytes())`, and on the pair itself rather than its hash. Encoding the shape rules out the 2×N/N×2 collision, and using the tuple as the key (not `hash(...)`) also closes the generic hash-collision nit for this cache — distinct matrices can never alias to one entry. In the real tree only square games reach the LP, but `test_non_square_via_lp` exercises the non-square path, so this is worth closing. Bonus: the tuple key is measurably *faster* than `hash(bytes)` (one hash of the tuple instead of `hash()` + a re-hash of the int), which widened the timing margin — Scotland k=3 strategy phase is now ~27.5 s.

**Should-fix — aliased cached strategy lists** — Documented rather than copied. Consumers are read-only today (`get_game_strategy` builds fresh rounded lists and never mutates the returned ones), so copying on every hit would be pure overhead on the hot path. Added a comment at the return site recording the read-only invariant so a future in-place mutation is an obvious violation.

**Nit — deprecated `.tostring()`** — Fixed. `get_game_strategy` now uses `hash(game_array.tobytes())` (identical bytes, drops the `DeprecationWarning`, and matches the new cache code).

**Nit — hash-collision re-check** — Closed for the new cache (tuple key, above). The per-stage caches in `games.py` keep the pre-existing `hash(...)` pattern; out of scope for this PR.

**Nit — thin <30 s margin** — The cache-key change also improved timing; now a consistent ~27.5 s (was ~28.5–29.2 s) on the reference machine.

No behaviour change: golden value still bit-identical (`5.875946490218083`), fast suite 70 passed.

Issue #26, opened 2026-07-04
User data directory for match folders and caches (pipx data gap)


svknoe:
A pipx/pip-installed user currently has no sane way to add their own match folders: `utilities.get_path` resolves `drafter/resources/matches/` relative to the installed package (site-packages), so user matrices must be edited inside the installed package — undiscoverable and wiped on upgrade — and solver caches (100s of MB with write_gamestates=True) are written into site-packages as well. Installed copies can effectively only draft against the bundled sample fixtures.

Fix: resolve match folders from a per-user data directory (e.g. platformdirs: `%APPDATA%/wtc-draft-solver/matches` on Windows) searched before the packaged samples, and write caches to the platform cache dir (if caches still exist by then). Packaged fixtures stay as read-only samples.

Slotting: belongs in the B2 solver-context refactor (#13), where path resolution moves out of the module globals anyway — do it there or immediately after. The web app (M4) makes this moot for end users, but the CLI keeps working this way for team power-users.

Found during the M1 review (PR #24 discussion in chat, 2026-07-04).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #13, opened 2026-07-03
B2: integer state encoding + solver context object


svknoe:
Replace verbose string keys (built by get_key, parsed back by regex) with packed integers: a team permutation = (defender, attacker_A, attacker_B, discarded, remaining-bitmask) over ≤8 players → one 32-bit int; a gamestate = two ints. Names exist only at the presentation layer.

Same PR-series retires the module-level globals (match_info, mutable settings) in favour of an explicit solver-context object — the from-scratch rebuild of the 2024 'store' idea (PLAN.md).

Acceptance: golden values unchanged; Scotland k=3 peak RAM < 200 MB (baseline ~2 GB); enumeration phase ≤ baseline 16 s.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

svknoe:
Related: #26 (user data directory for match folders/caches) — path resolution should come out of the module globals as part of this refactor; that issue covers where user data should live for installed (pipx) copies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

svknoe:
## B2 status (encoding + context landed; RAM criterion needs B3)

The B2 work landed in two PRs, both behaviour-preserving (solved values bit-identical, verified by direct comparison against `main`):

- **#35** — explicit `SolverContext`; retired the `settings`/`match_info` module globals and the hidden per-run caches.
- **#36** — packed 24-bit integer gamestate keys (`drafter/common/packing.py`); the engine works in name-sorted player indices, names only at the presentation layer; cache format v5. (`#37` then moved match/cache dirs off site-packages, closing #26.)

### Acceptance criteria
- ✅ **Golden values unchanged** — bit-identical to the pre-B2 engine (Smoke/Six vs `main`; Scotland k=3 slow test passes).
- ✅ **Enumeration phase ≤ 16 s** — measured **12.1 s** at Scotland k=3.
- ❌ **Peak RAM < 200 MB** — measured **~1816 MB** at Scotland k=3 (baseline ~2 GB).

### Why RAM isn't there yet — B3 is the lever
[...]

PR #35, opened 2026-07-05
B2a: explicit SolverContext, retire module globals (#13)

svknoe:
First of three PRs for **#13 (B2)** + **#26**, per the agreed plan
(context object → integer keys → user data dir). This is **PR-A**: retire the
mutable module-level globals in favour of an explicit `SolverContext`, threaded
through the solve and draft pipeline. **String gamestate keys are unchanged**,
so it is behaviour-preserving — the integer encoding lands in PR-B.

Part of #13.

## What changed

- **`SolverConfig`** (frozen) + **`SolverContext`** (`drafter/solver/context.py`):
  one object holding config, `enemy_team_name`, pairing tables, restriction data,
  and the gamestate / strategy / game-solution stores. Created once in
  `initialise()` and passed explicitly everywhere.
- **`PairingTables`** (`drafter/common/pairing.py`): owns the
  defender-picks-the-map value lookups, replacing `utilities.get_pairing_value`
  / `get_pairing_string` / `get_neutral_value` and the `match_info` pairing dicts.
- **`team_permutation`**: `RestrictionData` + `build_restriction` replace the
  module-level restrict-attackers caches and `enable_restricted_attackers`; the
  k-heuristic now takes an explicit `Side` instead of sniffing which pairing dict
  a name appears in.
- **`utilities.get_path(enemy_team_name, filename)`** — no longer reads a global.
- **`initialise_dictionaries.initialise(enemy_team_name, config) -> SolverContext`**;
  `read_pairing_matrix` returns a dict; `validate_best_not_below_worst(best, worst)`.
- Deleted `drafter/data/settings.py` and `drafter/data/match_info.py`.
- Harness (`smoke_draft`, `compute_golden_value`, `_solve_fixture`) and the
  in-process unit tests build a `SolverContext` instead of monkeypatching globals.

### Deliberate scoping note
The LP **normalised-matrix cache** in `utilities.py` (added by B1) stays
module-level: it is pure, content-keyed memoisation (same matrix → same
solution, cross-run-safe), not per-run/per-match state, and `get_game_solution(a)`
is a one-arg public API used by `test_zero_sum_solver.py`. The per-match mutable
state #13 targets is fully retired.

## Verification

- `pytest` (fast suite): **70 passed, 1 deselected** in ~3 s.
- `pytest -m slow` (Scotland k=3): **1 passed** in ~44 s — value
  `5.875946490218083`, supports `{Mariusz}` / `{Drukhari}`, **bit-identical** to
  the pinned golden values.
- `scripts/smoke_draft.py` end-to-end: completes (Smoke total 11.0, expected 5.0),
  identical to before.

Golden values are structurally unchanged: key construction and value lookup are
untouched, so the payoff matrices fed to the solver are identical.

> Correction carried in this PR: the live Scotland k=3 golden value is
> `5.875946490218083` (`test_golden_values.py:136`), not the `6.189614…` that had
> lingered in a docstring — fixed here.


svknoe:
## Reviewer-agent review (PR-A)

Reviewed the full `main...b2a-solver-context` diff (25 files, +532/-500) with the primary lens the PR asks for: **is this behaviour-preserving?** I read the new `context.py` / `pairing.py`, diffed every threaded solver/draft/data module against `main`, and checked the k-restriction `Side` refactor, the discard child-key wiring, the factories, and every test change. I also re-ran the fast suite locally: **70 passed, 1 deselected in ~3s**.

**Verdict: this looks solid and I believe it is behaviour-preserving. No blocking issues found.** Details below, then a couple of nits.

### Correctness checks that passed

- **Context threading is complete.** A grep for live reads of `match_info.*`, `settings.<knob>`, `get_pairing_value/_string/_neutral_value`, `enable_restricted_attackers`, and the module-level `dictionaries` across `drafter/` and `scripts/` finds only comments/docstrings — no live global reads remain. `drafter/data/settings.py` and `drafter/data/match_info.py` are deleted.
- **`SolverConfig` defaults are byte-for-byte the old `settings.py` values** (`friendly_team_name="Norway"`, all `show_*`/`invert_discard_attackers`/`require_unique_names=True`, `neutral_map_weight=0.5`, all read/write toggles `True`, `restrict_attackers=True`, `restricted_attackers_count=4`). So `SolverConfig()` in `app.py`/`smoke_draft.py` reproduces the previous default run exactly.
- **k-restriction `Side` refactor is correct** (`team_permutation.get_heuristically_best_attackers`): `FRIENDLY` → `friendly_vs_defender`/`friendly_vs_field` + `ranking_sign = -1`; `ENEMY` → `enemy_vs_*` + `ranking_sign = 1`. That matches the original (`ranking_sign = -1`, then `*= -1` in the enemy branch). `build_restriction` maps the four dicts identically to the old `enable_restricted_attackers` (`friendly_vs_defender=worst`, `enemy_vs_defender=transpose(best)`, fields = neutral / transpose(neutral)). The explicit `Side` also removes a latent fragility in the old "sniff which dict the name is in" trick.
- **`game_state.get_next_gamestate_matrix`** passes `Side.FRIENDLY` for the friendly permutations and `Side.ENEMY` for the enemy ones, with the parent/opposing permutation arguments in the same order as before.
- **`PairingTables.value`** reproduces `get_pairing_value` exactly (friendly-defender→best, enemy-defender→worst, none→`worst + neutral_weight*(best-worst)`, unknown-defender→`ValueError`). `neutral_dictionary()` matches the old `get_neutral_pairing_dictionary`.
- **discard child-key reconstruction (`games.discard_attacker`)** is unchanged (same `GameState(DraftStage.none, ...)` key construction; refused-pair wiring intact) and now reads `select_defender_strategies` / `ctx.pairing.value` correctly. `test_discard_wiring` pins the n=6 matrix cell-by-cell and passes.
- **Factories match the old module-level initialisers.** `make_gamestate_dictionaries` yields the same 11 named empty dicts; `make_strategy_dictionaries` yields the same 9 `{'descriptor': [n, stage]}` seeds; `make_game_solution_caches` gives one `{}` per `(stage, n)` — same shape the old `select_/discard_/attackers_cache[n]` module dicts had, now addressed as `ctx.game_solution_caches["<stage>"][n]`.
- **Tests preserve coverage.** `test_map_model` / `test_input_validation` now drive `PairingTables.value` and `read_pairing_matrix(path, require_unique_names)` directly, asserting the same normalisation, defender-map, and validation behaviour, and even *add* a case (`test_duplicate_name_across_teams_allowed_when_setting_disabled`). `test_discard_wiring` builds a real `SolverContext` and stubs `utilities.get_game_strategy` (correct — `games.py` calls it via the module reference). Harness scripts convert monkeypatch-globals → `SolverConfig`.

### On the deliberately-kept module global

`normalised_game_solution_cache` in `utilities.py` staying module-level is **acceptable for this PR**. It is keyed on the `(shape, tobytes)` of the positivity-shifted matrix and a zero-sum game's value/strategies are translation-invariant, so it is genuinely pure, content-addressed memoisation — same input always yields the same output, no per-match/per-run state. It cannot leak match-specific state between runs, and `get_game_solution(a)` remaining a 1-arg API keeps `test_zero_sum_solver.py` working. The PR is explicit about this scoping, and the golden tests still run each solve in a fresh subprocess, so nothing observable changes.

### Nits (non-blocking)

1. **Stale docstring references to deleted modules in `scripts/brute_force_oracle.py`** (lines 14, 25): still mention `settings.neutral_map_weight` / `settings.restrict_attackers`. That file was not touched by this PR and is a standalone oracle (no import of the deleted modules, so it still runs), but the references are now dangling. Worth a one-line docstring touch-up in a follow-up.
2. **Pre-existing unused import** `from enum import Enum` in `drafter/common/utilities.py` (line 2) — present on `main` too, not introduced here. Since this PR is already tidying that module, dropping it would be a free cleanup, but it's out of scope.

Nice touch: `set_enemy_team.prompt_enemy_team` now `return`s the recursive re-prompt result, fixing a latent drop-the-value bug in the old confirm==False path.

PR #36, opened 2026-07-05
B2b: packed-integer gamestate keys, index-based engine (#13)

svknoe:
Second of three PRs for **#13 (B2)** + **#26** (context → **integer keys** → user data dir). Builds on PR-A (#35). Part of #13.

Replaces the verbose string gamestate/team-permutation keys with **packed 24-bit integers**. The engine now works in name-sorted player indices; names live only at the presentation layer.

## What changed
- **`drafter/common/packing.py`**: a team permutation packs `(remaining bitmask | 4×4-bit role slots)` into one int; a gamestate key is the `(friendly, enemy)` code pair; draft stage is derived from the code. Property-tested in `test_packing.py` (round-trip, injectivity, `n`/stage agreement).
- **`PairingTables`** → index-keyed **numpy arrays** + a `Defender` enum (friendly-defends → best, enemy → worst, none → neutral midpoint).
- **`NameIndex`** (context.py) assigns indices in **name-sorted order** — deliberately matching the old string-key engine's ordering everywhere — so this is a pure relabelling: enumeration order, payoff matrices and equilibrium selection are unchanged.
- **`TeamPermutation`/`GameState`** hold indices; `get_key` packs, `get_..._from_key` decodes.
- **`draft.py`** maps indices↔names at the UI boundary (display, input matching, pairings); a discard-stage subtlety is handled: each team chooses among the *opponent's* attackers, so that choice space is the opponent's index space.
- **Cache format v5**: int-tuple keys serialised as `[f,e]` lists; strategy dicts as `[key, strategy]` pair lists (JSON can't key on tuples); the marker now records the player-name ordering and rejects a cache whose CSV names/order changed.

## Verification
- **Bit-identical values vs `main` (PR-A)** — verified by direct comparison: Smoke `4.997553182223154`, Six `1.9058104884743532` (both match `main` exactly). Scotland k=3 slow test passes (value + supports). Smoke draft output identical.
- Fast suite **80 passed** (adds packing + `from_dicts` tests); v5 cache round-trip verified (write→read bit-identical).
- **Enumeration phase 12.06 s** ≤ 16 s baseline. ✓

> Note: the golden pins (`...153`, `...512`) are 1–2 ulp off from what **both** `main` and this branch produce (`...154`, `...532`) — a pre-existing B1-LP-solver artifact covered by the tests' `abs=1e-9`. Not touched here; this PR changes nothing vs PR-A.

## RAM (acceptance criterion) — deferred to B3
Scotland k=3 peak working set is **~1816 MB** (baseline ~2 GB), which **misses the <200 MB target**. As PLAN.md anticipated, packed keys shrink the *keys*, but the ~1.5M `GameState` objects and ~950k strategy vectors stored as dict *values* dominate RAM; the <200 MB win requires **B3 (value-only backward induction)** — "kills the 271 MB strategy dictionaries". Per the agreed plan ("measure after B2, defer B3 if needed"), B3 is a separate follow-up. This PR lands the encoding foundation B3 builds on.

### DIFF
--- a/drafter/common/game_state.py
+++ b/drafter/common/game_state.py
@@ -8,9 +8,11 @@ class GameState:
         self.friendly_team_permutation = friendly_team_permutation
         self.enemy_team_permutation = enemy_team_permutation
 
-    def get_key(self, leading_whitespace = ""):
-        return "{}Friends: {}\n{}Enemies: {}".format(leading_whitespace, self.friendly_team_permutation.get_key(), 
-            leading_whitespace, self.enemy_team_permutation.get_key())
+    def get_key(self):
+        # A gamestate key is the pair of packed-integer team-permutation codes
+        # (GitHub issue #13, B2). Hashable and compact; names are resolved only
+        # at the presentation layer.
+        return (self.friendly_team_permutation.get_key(), self.enemy_team_permutation.get_key())
 
     def get_n(self):
         friendly_n = self.friendly_team_permutation.get_n()
@@ -67,9 +69,9 @@ def get_next_gamestate_matrix(ctx, gamestate):
 
 
 def get_gamestate_from_key(key):
-    friendly_team_representation, enemy_team_representation = key.split('\n', 1)
-    friendly_team_permutation = team_permutation.get_team_permutation_from_key(friendly_team_representation)
-    enemy_team_permutation = team_permutation.get_team_permutation_from_key(enemy_team_representation)
+    friendly_code, enemy_code = key
+    friendly_team_permutation = team_permutation.get_team_permutation_from_key(friendly_code)
+    enemy_team_permutation = team_permutation.get_team_permutation_from_key(enemy_code)
     draft_stage = friendly_team_permutation.get_draft_stage()
 
     return GameState(draft_stage, friendly_team_permutation, enemy_team_permutation)
--- a/drafter/common/packing.py
+++ b/drafter/common/packing.py
@@ -1,0 +1,96 @@
+"""Packed-integer encoding of team permutations and gamestates (GitHub issue
+#13, B2). Replaces the verbose human-readable string keys.
+
+A team permutation over <=8 players packs into a 24-bit int:
+
+    bits  0..7  : remaining-players bitmask (bit i set => player i remains)
+    bits  8..11 : defender           (4 bits, ROLE_NONE = 0xF when unset)
+    bits 12..15 : attacker_A         (4 bits)
+    bits 16..19 : attacker_B         (4 bits)
+    bits 20..23 : discarded_attacker (4 bits)
+
+Indices are per side (friendly = CSV rows, enemy = CSV columns), 0..n-1. A
+gamestate key is the pair (friendly_code, enemy_code); the draft stage is
+derivable from either code (see draft_stage_of_code). Names live only in the
+presentation layer (drafter.solver.context NameIndex).
+"""
+from drafter.common.draft_stage import DraftStage
+
+ROLE_NONE = 0xF
+
+_DEFENDER_SHIFT = 8
+_ATTACKER_A_SHIFT = 12
+_ATTACKER_B_SHIFT = 16
+_DISCARDED_SHIFT = 20
+_ROLE_MASK = 0xF
+_REMAINING_MASK = 0xFF
+
+
+def _pack_role(index):
+    return ROLE_NONE if index is None else index
+
+
+def _unpack_role(code, shift):
+    value = (code >> shift) & _ROLE_MASK
+    return None if value == ROLE_NONE else value
+
+
+def mask_from_indices(indices):
+    mask = 0
+    for index in indices:
+        mask |= (1 << index)
+    return mask
+
+
+def indices_from_mask(mask):
+    # Ascending index order -- reproduces the old "remaining players sorted"
+    # canonicalisation for free (a bitmask is inherently order-free).
+    return [i for i in range(8) if mask & (1 << i)]
+
+
+def encode_team_permutation(remaining_mask, defender, attacker_a, attacker_b, discarded):
+    return (remaining_mask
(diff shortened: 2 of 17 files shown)


svknoe:
## Reviewer-agent review (PR-B)

**Verdict: approve.** The behaviour-preserving claim holds under independent verification, the encoding is sound and well property-tested, and the fiddliest part (draft.py index spaces) is correct end-to-end. No blocking issues. Three cosmetic nits below.

### Verification I ran
- **Fast suite: 80 passed, 1 deselected** (matches the PR body).
- **Bit-identical values independently confirmed against a fresh `main` worktree:** Smoke `4.997553182223154`, Six `1.9058104884743532` on both branches. The pure-relabelling claim is real.
- **Cache v5 round-trip inspected on disk.** The `strategy_4_player_select_defender` cache serialises as `[[16776975, 16776975], [[[0, 0.293], [1, 0.0], [2, 0.691], [3, 0.015]], [...], 4.997553182223154]]` — packed-int key list, index-keyed strategy entries, full-precision value. Re-tupling `{tuple(key): strategy for key, strategy in serialised}` reproduces a hashable key and the correct value. Marker: `{"cache_format_version": 5, "friendly_names": [...], "enemy_names": [...]}`.
- **Smoke draft to completion.** The discard stage renders in the correct index spaces — `Norway options: Carol vs - Eldar, Ork` (friendly refuses among *enemy* attackers) and `Smoke options: Tau vs - Bob, Dave` (enemy refuses among *friendly* attackers). Final pairings put `(D)` on the correct defending side across both spaces (`Carol (D) vs Tau`, `Dave vs Ork (D)`). No wrong-space names, no crash.

### What I checked carefully and found correct

**Encoding (`packing.py`).** Disjoint bit fields (mask 0–7, four 4-bit slots at 8/12/16/20), so `encode_team_permutation` is a bijection on the field tuple; two distinct reachable permutations cannot collide. Role indices are always 0–7 (n ≤ 8) and `ROLE_NONE=0xF` all fit in 4 bits, mask bits ≤ 7 fit the byte — no overflow into an adjacent slot is reachable. `team_permutation_n` (popcount + roles) and `draft_stage_of_code` mirror the old `get_n`/`get_draft_stage`. `test_packing.py` covers round-trip, injectivity, 24-bit fit, stage and n — good coverage.

**Pure-relabelling claim.** The load-bearing argument: old `TeamPermutation.__init__` did `sorted()` on *name strings* (lexicographic); new does `sorted()` on *name-sorted indices* (numeric) — identical order because `NameIndex.from_names` sorts. `get_initial_game_state` switched from `best.keys()` / `best[friends[0]].keys()` (CSV order) to `range(n)` (name-sorted), but since both feed through the same `sorted()`, enumeration order is unchanged. `from_dicts` places each `(friendly_name, enemy_name)` by name-sorted index and is independent of CSV column/row order (covered by `test_pairing_tables_from_dicts_maps_names_to_name_sorted_indices` with deliberately-unsorted CSV order). The CSV reader builds every friendly row against one shared `enemies` header, so `next(iter(best.values())).keys()` reliably yields the full enemy set.

**PairingTables.value.** numpy `[i, j]` indexing reproduces the old dict semantics exactly: FRIENDLY→best, ENEMY→worst, none→`worst + w*(best-worst)`. No transposition (rows friendly, cols enemy throughout). `float(...)` casts keep the return a Python float. The `RestrictionData` transposes are correct: `enemy_vs_defender = best.T` gives `[enemy_attacker, friendly_defender] = best[friendly_defender][enemy_attacker]`, matching the old `get_transposed_pairing_dictionary(best)` exactly; likewise `enemy_vs_field = neutral_matrix.T`.

**draft.py index spaces.** In the discard stage the friendly strategy holds *enemy* indices and vice-versa (games.py passes `[e_attacker_A, e_attacker_B]` as the friendly row options). `print_team_options` sets `choice_name_index = opponent_name_index` for discard and own space otherwise; `get_next_gamestate` does the cross-space swap (`f_discarded = ctx.friendly.index(enemy_team_selection)`) with an accurate comment. The `invert_discard_attackers` swaps (suggestion + user selection) operate purely in name space, preserving old semantics.

**Cache guard.** Marker now also stores/validates `friendly_names`/`enemy_names`, so a cache whose CSV names or order changed is rejected — correct, since the packed keys are positional. `CACHE_FORMAT_VERSION` bumped 4→5. Gamestate cache re-tupled on read; `get_gamestate_from_key` unpacks a 2-element list or tuple interchangeably.

**RAM deferral to B3** is reasonable and matches PLAN.md ("measure after B2, defer B3"). The ~1816 MB is dominated by the `GameState` object values and strategy vectors, which are exactly what B3 (value-only backward induction) removes; packed keys shrink only the keys. One cheap thing left on the table, noted below as a nit, not a blocker.

### Nits (non-blocking)

1. **Dead code in `utilities.py`.** `get_transposed_pairing_dictionary` (line 7) and `get_value_from_input_dictionary` (line 275) are now unreferenced anywhere in the repo (the transpose moved to `.T`, the lookup moved into numpy indexing). Safe to delete in this PR or a cleanup follow-up.

2. **Stale doc comment, `context.py:63-64`.** The `SolverContext` docstring still says *"Names still live in the key strings at this step; the packed-integer encoding (B2 second PR) moves the name<->index maps here too."* — this *is* the second PR and the maps are now here (`friendly`/`enemy` fields). Update to reflect the achieved state.

3. **`game_state_dictionaries.write_dictionaries:62`** still names its local `string_representation` though it now holds int-tuple keys, and (per the PR's own RAM note) the gamestate dict still stores full `GameState` objects as values rather than decoding on demand from keys. Neither affects correctness; the rename is trivial and the value-storage question is legitimately B3's.

Nice touch: `test_discard_wiring.py` and `test_map_model.py` were migrated to index/numpy fixtures rather than dropped, and a `from_dicts` name-ordering test was added — the exact places a relabelling bug would hide.

PR #37, opened 2026-07-05
B2c: per-user match & cache directories via platformdirs (#26)

### LINKED ISSUE #26: User data directory for match folders and caches (pipx data gap)

svknoe:
Third of three PRs for **#13 (B2)** + **#26** (context → integer keys → **user data dir**). Builds on PR-A (#35) and PR-B (#36). Closes #26.

Today a pip/pipx user can only draft against the bundled sample matrices: `get_path` resolved match folders **and** wrote 100s-of-MB caches inside the installed package (site-packages) — undiscoverable and wiped on upgrade.

## What changed
- **`drafter/data/paths.py`**: match inputs resolve from a **per-user data dir first, packaged samples second** (`platformdirs.user_data_dir` → `%APPDATA%\wtc-draft-solver\matches` on Windows); solver caches go to the **platform cache dir** (`user_cache_dir`), created on demand. `MatchPaths(input_dir, cache_dir)` is carried on the `SolverContext` (`ctx.paths`).
- **`set_enemy_team`** lists `paths.list_available_teams()` — user + packaged, deduped (a user folder shadows a sample of the same name).
- `initialise` / `game_state_dictionaries` / `strategy_dictionaries` read+write via `ctx.paths.input_file` / `cache_file`; `read_write` write helpers `mkdir` the cache dir.
- Removed the now-unused `utilities.get_path` (and its `Path` import). Packaged `drafter/resources/matches/` folders are read-only samples.
- `platformdirs==4.10.0` added to `pyproject.toml` (packaging already ships only CSVs, so caches never enter the wheel).

## Verification
- Golden tests fall back to the packaged samples (no user dir set), so values are **unchanged** — fast suite **86 passed** (adds `test_paths.py`: packaged fallback, user-first, dedup, cache-dir auto-create), smoke draft output identical.
- Verified a real **write→read cache round-trip in the platform cache dir** (`%LOCALAPPDATA%\wtc-draft-solver\Cache\matches\Smoke`): 15 files written, dir auto-created, value bit-identical, cleaned up.
- Docs updated (README: where to put your own match folders per-OS; CLAUDE.md: paths.py + cache location).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -36,7 +36,8 @@ drafter/
     pairing.py             PairingTables: defender-picks-the-map value lookups for one match
   data/
     initialise_dictionaries.py  reads CSVs (read_pairing_matrix), builds the SolverContext, solves
-    set_enemy_team.py      InquirerPy menu over drafter/resources/matches/<team>/ (returns the name)
+    paths.py               match-folder + cache resolution (platformdirs, issue #26)
+    set_enemy_team.py      InquirerPy menu over paths.list_available_teams() (returns the name)
     read_write.py          JSON cache IO
   solver/
     context.py             SolverConfig (the knobs) + SolverContext (all per-run state, passed explicitly)
@@ -45,12 +46,20 @@ drafter/
     games.py               build the payoff matrix for one gamestate from child values
     draft.py               interactive draft loop (uses plain input(), not InquirerPy)
     draft_loop.py          replay wrapper
-  resources/matches/<Team>/    one folder per opponent = the "database"
+  resources/matches/<Team>/    packaged read-only SAMPLE opponents (issue #26)
     pairing_matrix_best.csv    required input (see format below)
     pairing_matrix_worst.csv   required input
-    cache_format.json          cache version marker (see read_write.py)
-    *_dictionary.json          cached preprocessing output (gitignored, can be 100s of MB)
 ```
+
+Match folders resolve from a **per-user data dir first, packaged samples
+second** (`drafter/data/paths.py`, issue #26): a user's own opponents live in
+`platformdirs.user_data_dir("wtc-draft-solver")/matches/<Team>/`
+(`%APPDATA%\wtc-draft-solver\matches` on Windows), searched before the bundled
+samples under `drafter/resources/matches/`. Solver JSON caches
+(`cache_format.json`, `*_dictionary.json`) are written to the platform **cache**
+dir (`user_cache_dir(...)/matches/<Team>/`), never into the package — so a
+pip/pipx install stays clean. `ctx.paths` (a `MatchPaths`) carries `input_dir`
+and `cache_dir` for the current match.
 
 ## How it works (important for any performance work)
 
@@ -182,7 +191,8 @@ Notes for agents:
   test needs `-m slow`, ~5 min). CI (`.github/workflows/ci.yml`) runs the
   smoke test + fast suite on every push/PR — keep it green; run `pytest`
   before pushing solver changes.
-- JSON caches under `drafter/resources/matches/` are gitignored; never commit
-  them. Never commit `.venv/` or `__pycache__/`.
+- Solver JSON caches now live in the platform cache dir (issue #26), not the
+  repo; any stray `*.json` under `drafter/resources/matches/` is gitignored —
+  never commit caches, `.venv/`, or `__pycache__/`.
 - PLAN.md holds the current revival roadmap (performance, 11th-edition map
   rules, web distribution). Read it before starting substantial work.
--- a/README.md
+++ b/README.md
@@ -25,6 +25,21 @@ matchup (10–10). Best must be ≥ worst in every cell.
 Folders in the old single-matrix format can be converted with
 `python scripts/migrate_match_folder.py drafter/resources/matches/<Team>`.
 
+### Where match folders live
+
+The bundled `drafter/resources/matches/<Team>/` folders are **read-only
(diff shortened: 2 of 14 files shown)


svknoe:
## Reviewer-agent review (PR-C)

Reviewed the full diff (`git diff main...b2c-user-data-dir`), read every touched module in context, grepped for dangling refs, and ran the suite + smoke draft locally.

**Verdict: LGTM, no blocking issues.** The path indirection is clean, behaviour-preserving, and well-tested. A few non-blocking observations below.

### Verification I ran
- `pytest -q` -> **86 passed, 1 deselected** (matches the PR claim).
- `platformdirs==4.10.0` installed; `user_data_dir('wtc-draft-solver', appauthor=False, roaming=True)` -> `%APPDATA%\wtc-draft-solver`, `user_cache_dir(..., appauthor=False)` -> `%LOCALAPPDATA%\wtc-draft-solver\Cache`. Args behave as documented.
- `grep get_path` -> **no matches** anywhere. `get_path` and its `Path` import are gone from `utilities.py` with zero dangling references; no other module hardcodes `resources/matches` at runtime (remaining hits are docs/CLI-arg scripts).
- Smoke draft (`scripts/smoke_draft.py Smoke`) solved and entered the draft loop via the new `resolve_match` fallback; no stray JSON written under `drafter/resources/matches`.
- `SolverContext` is only ever built by keyword (`initialise` + `test_discard_wiring`), so inserting the `paths` field mid-dataclass is safe. `test_discard_wiring` correctly passes `paths=None`.

### What's solid
- **Write-path coverage is complete.** All three cache writers `mkdir(parents=True, exist_ok=True)`: `write_dictionary` (gamestate + strategy dicts both route through it) and `write_cache_format_marker` (the marker). `exist_ok=True` is correct for repeated runs.
- **Reads degrade gracefully.** `read_dictionary` returns `None` on any failure and `cache_format_is_current` catches `OSError`, so a nonexistent cache dir never crashes a fresh solve.
- **Behaviour preservation holds.** Golden/smoke fixtures set no user dir, so `resolve_match` falls back to the packaged sample; golden solves additionally disable cache read/write entirely. Values are untouched.
- **Missing-team case is safe** (not just menu-guarded): `resolve_match` on an absent team yields a packaged path that doesn't exist, and `read_pairing_matrix` raises a clear `InputError('Missing input file: ...')` rather than an obscure crash.
- **Packaging stays clean:** `package-data` still ships only `*.csv`/`*.txt`, so caches can never enter the wheel; `platformdirs` pinned consistently with the rest.

### Non-blocking observations

**1. (nit) Tests write to the *real* user cache dir if a test ever enables cache writes.**
`test_paths.py` monkeypatches `user_matches_dir`/`user_cache_root` only within its own tests. That's fine for the current suite (golden tests disable writes; `test_write_dictionary_creates_missing_cache_dir` passes an explicit `tmp_path`). But there's no session-wide guard redirecting `user_cache_root` to a temp dir, so any *future* test that runs `initialise` with `write_*=True` would silently populate `%LOCALAPPDATA%\wtc-draft-solver\Cache\matches` on the dev/CI machine. Consider an autouse fixture in `conftest.py` that points `user_cache_root` (and `user_matches_dir`) at `tmp_path` for the whole session as cheap insurance. Not required for this PR.

**2. (nit) Test gap: the ''user dir exists but lacks *this* team'' fallback.**
`test_resolve_match_falls_back_to_packaged` exercises the case where the user *matches* dir doesn't exist at all. It does not cover the case where `user_matches_dir()` **exists** but has no `<team>` subfolder — the exact path most upgraded users hit (they've created the matches dir for other opponents). The code handles it correctly (`user_dir.is_dir()` is per-team), but a one-line test (create `user/matches/` empty, resolve `Smoke`, assert it still lands on the packaged sample) would lock in that behaviour.

**3. (minor, worth a sanity check) `roaming=True` for user data.**
Match CSVs going in the *roaming* profile means they sync across a user's domain machines — probably the friendly intent (your rosters follow you). Just flagging it's a deliberate choice: on managed/enterprise Windows this inflates the roaming profile and the folder differs from the non-roaming `%LOCALAPPDATA%`. The README/CLAUDE.md correctly document `%APPDATA%`, so docs and code agree. No change needed unless you'd rather keep inputs local.

**4. (pre-existing, not this PR) bare `except:` in `read_dictionary`.**
`read_write.read_dictionary` uses a bare `except:` that also swallows `KeyboardInterrupt`/`SystemExit`. Untouched by this PR; noting only because this PR leans on it for the graceful-read guarantee. Fine to leave.

Nice, tightly-scoped PR that finishes the #26 story cleanly.

-- Reviewer agent (Opus 4.8)

PR #39, opened 2026-07-05
B3: value-only backward induction, ~120 MB at Scotland k=3 (#13)

svknoe:
The RAM step for **#13**. Builds on B2a (#35) / B2b (#36). With B2, Scotland k=3 peaked at ~1816 MB (the packed keys shrank the *keys*, but the ~1.5M `GameState` objects and ~950k strategy vectors stored as dict *values* dominated). B3 stops storing those.

**Result: Scotland k=3 peak working set ~120 MB (down from ~1.8 GB, ~15×) — meets #13's <200 MB criterion.**

## What changed
- **Single-int gamestate keys** (`packing.py`): `friendly | enemy<<24`, so a key fits a numpy int64.
- **Enumeration** (`game_state_dictionaries.enumerate_gamestates`): BFS storing each `(n, stage)` level as a **sorted numpy int64 key array** — no `GameState` objects kept (decoded on demand). The whole tree ≈ 12 MB.
- **Value-only backward induction** (`strategy_dictionaries.initialise_values`): each gamestate's game is built from its children's values (`games.build_game`) and solved for the **value only** (`utilities.get_game_value`). Values live per stage as parallel sorted `(keys, values)` arrays (`StageValues`, binary-search lookup) — ≈ 8 MB vs ~810 MB of strategy vectors. Only the LP solves stay memoised; the per-matrix value cache is dropped (re-solving small games is cheap; caching them cost ~70 MB of peak).
- **Draft** recomputes the labelled strategy for each visited gamestate on demand from child values; the same recursion (`game_value`) solves off-tree, non-k-restricted moves — replacing the old tree-extension machinery.
- **No disk cache**: its JSON format was incompatible with the value arrays and loading it would spend the RAM we're saving. The solve is fast (~28 s) and held in-process for the whole session. #26's user-matches-dir *input* resolution is unchanged; only the (now-unused) cache-dir sub-feature is superseded.
- Removed now-dead helpers (dictionary-name / arbitrary-entry) and the cache config flags.

## Verification
- **Bit-identical values vs `main`**: Smoke `4.997553182223154`, Six `1.9058104884743532`, Scotland k=3 slow test passes (value `5.875946490218083`, supports `{Mariusz}`/`{Drukhari}`); smoke draft output identical.
- **RAM (Scotland k=3, peak working set), 3 runs: 116.6 / 115.9 / 117.2 MB** (was 1816 MB).
- **Enumeration ~8 s** ≤ 16 s; full solve ~28 s.
- Fast suite **88 passed** (adds a gamestate-key round-trip test; `test_discard_wiring` now checks `build_discard_game` directly).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,11 +38,11 @@ drafter/
     initialise_dictionaries.py  reads CSVs (read_pairing_matrix), builds the SolverContext, solves
     paths.py               match-folder + cache resolution (platformdirs, issue #26)
     set_enemy_team.py      InquirerPy menu over paths.list_available_teams() (returns the name)
-    read_write.py          JSON cache IO
+    read_write.py          JSON IO (used only by scripts/migrate_match_folder.py now)
   solver/
     context.py             SolverConfig (the knobs) + SolverContext (all per-run state, passed explicitly)
-    game_state_dictionaries.py  enumerate all reachable gamestates per (n, stage)
-    strategy_dictionaries.py    backward induction: solve every gamestate's game, deepest first
+    game_state_dictionaries.py  enumerate reachable gamestate keys per (n, stage) -> numpy int64 arrays
+    strategy_dictionaries.py    value-only backward induction; StageValues + draft strategy recompute (B3)
     games.py               build the payoff matrix for one gamestate from child values
     draft.py               interactive draft loop (uses plain input(), not InquirerPy)
     draft_loop.py          replay wrapper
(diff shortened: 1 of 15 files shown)


svknoe:
## Reviewer-agent review (B3)

Value-only backward induction, #39 (`b3-value-only`). I read the full `main...b3-value-only` diff and independently verified the correctness-critical pieces (enumeration equivalence, child-stage mapping, off-tree recursion, key range) rather than trusting the reported numbers. **Verdict: I could not find a correctness bug. This is a clean, well-reasoned rewrite and I'd merge it.** No blocking issues; two minor/robustness notes and some praise below.

### What I verified (empirically, this branch)

- **Bit-identical values.** Smoke `4.997553182223154`, Six `1.9058104884743532` (fresh solves), and the slow Scotland k=3 test passes (value `5.875946490218083`, supports `{Mariusz}`/`{Drukhari}`) in ~29 s. 88 fast tests pass; `-m slow` passes.
- **Enumeration is provably equivalent to the old set-based BFS.** I reconstructed the old recursive enumeration (Python-set dedup, traversing *all* stages incl. discard, stopping before `(4, discard)`) and compared it cell-for-cell to `enumerate_gamestates` output:
  - Six k=4 — every stored `(n, stage)` bucket equal: `(4,none)` 223, `(4,sd)` 3568, `(4,sa)` 32112, `(6,none)` 1, `(6,sd)` 36, `(6,sa)` 1296; discard levels traversed (5184 at n=6) but correctly **not** stored.
  - Smoke — `(4,none)` 1, `(4,sd)` 16, `(4,sa)` 144, all equal; exercises the `(4, discard)` stop condition.
  - All stored arrays are strictly sorted and unique (so `np.searchsorted` is valid), and no discard bucket is ever stored.
- **Child-stage mapping is right, and children precede parents.** `solve_order` = `[(4,sa),(4,sd),(4,none),(6,sa),(6,sd),(6,none),(8,sa),(8,sd),(8,none)]`. `child_stage` maps `none→(n,sd)`, `sd→(n,sa)`, `sa→(n-2,none)`. Cross-checking `build_game`'s dispatch (`none→select_defender game`, `sd→select_attackers game`, `sa→discard game`): each stage's children resolve to a bucket solved strictly earlier — `(6,sa)` reads `(4,none)` (pos 3 < 4), `(8,sa)` reads `(6,none)` (pos 6 < 7), same-n `none→sd→sa` are contiguous and ordered. Confirmed the `(n-2, none)` child of a `select_attackers` state really has `get_n()==n-2` (defender+2 attackers removed from an n-state = n-3 remaining, +1 returned refused attacker = n-2).
- **Every parent→child reference is present in its child store.** I walked every stored gamestate, collected the exact child keys `build_game` fetches, and confirmed all are present in the `child_stage` store (Six k=4: 0 missing, 0 that would be out-of-bounds in `value()`). So the unchecked `searchsorted` in `StageValues.value()` is always valid during backward induction.
- **Discard wiring unchanged from `main`.** The n≥6 off-diagonal terms return the *refused* pair `(f_B,e_A)`/`(f_A,e_B)` to the child `none` pools exactly as merged in #32 (`test_discard_wiring` now pins `build_discard_game` directly, cell-by-cell), and the n==4 closed form matches `main`'s `discard_attacker_4` line-for-line. Row/col labels preserved as `[e_A,e_B],[f_A,f_B]`.
- **Off-tree recursion terminates, memoises, and agrees.** Constructed a non-k-restricted `(6, select_attackers)` state (key absent from the stored array): `game_value` returns via the recursion, memoises into `extension_values`, and `game_strategy`'s value plus an independent `build_game`+`get_game_value` recompute all agree to 1e-12. Base case (4-player closed form, no getter call) reached correctly.
- **Single-int key fits int64.** Max 24-bit team codes → `encode_gamestate` = `0xFFFFFFFFFFFF` (48 bits, `< 2^63`); round-trips and survives `np.int64` (realistic max is 47 bits). `test_gamestate_key_round_trip` pins the 48-bit budget.
- **Draft output identical.** Drove `smoke_draft.py` with piped defaults on both `main` and this branch (same `random.seed(0)`): all 703 lines of interactive draft content are byte-identical; the only diff is the trailing traceback's file paths/line numbers (an artifact of my worktree path + running out of piped input, not a behavior change). `game_value(ctx, gamestate_tree[0])` produces the correct "Expected result" at finish.
- **No dead references / `read_write` not orphaned.** Grep for every removed symbol (`read_gamestates`/`write_*`, `game_solution_caches`, `make_*_dictionaries`, `get_arbitrary_dictionary_entry`, `get_*_dictionary_name`, `discard_attacker_4`, `get_game_array`, `extend_dictionary`, `get_dictionary_for_gamestate`, `update_dictionaries`, `process_gamestate_dictionary`) is clean. `read_write.py` is still used by `scripts/migrate_match_folder.py` and two tests.
- **Value-cache drop is value-neutral.** `get_game_value` is deterministic (`get_game_solution` → saddle / 2×2 closed form / one translation-normalised HiGHS LP, itself still deduped). Dropping per-matrix memoisation only affects speed. The shared `draft_strategy_cache` for `game_strategy` is content-addressed (`hash(tobytes())`), so sharing one cache across stages is safe.

### Minor / robustness (non-blocking)

1. **`StageValues.value()` has no bounds guard — silent wrong answer if ever called on an absent key** (`drafter/solver/strategy_dictionaries.py:39-40`). `value()` does `float(self.values[np.searchsorted(self.keys, key)])` with no `contains` check. I verified every current call site only reaches it for keys that are present (`known_value` guards with `contains()` first; backward induction only fetches enumerated children), so it's correct **today**. But it's a latent trap: a future refactor that calls `value()` for an absent key would silently read a *neighbouring* gamestate's value (or `IndexError` at the array end) instead of failing loudly. Consider `assert self.keys[index] == key` inside `value()` (cheap, once per lookup) so a future mis-wire surfaces as a crash, not a wrong equilibrium value. Optional.

2. **`_missing_child` message slightly under-describes its trigger** (`strategy_dictionaries.py:64-65`). It fires whenever `child_store is None`, which for the enumerated tree only happens at `(4, select_attackers)` (child `(2, none)` doesn't exist) — and there `build_discard_game`'s n==4 branch never calls the getter, so it correctly never raises. Matches the comment; no action needed.

### Praise

- The `solve_order` / `child_stage` / `build_game`-dispatch trio is a genuinely elegant restatement of the backward induction — the child-stage table makes the "children before parents" invariant obvious and checkable, exactly where you want it: the piece most likely to silently corrupt values.
- Collapsing enumeration to "each BFS level is one `(n, stage)`, `np.unique` it, don't store discard" is a clean simplification over the old recursive `perform_gamestate_tree_extension` / `get_previous_gamestate_dictionary` machinery, and the discard-passthrough reasoning is spot on.
- Unifying the old bespoke tree-extension path with the ordinary `game_value` recursion for off-tree moves removes a whole class of "extension re-solve" edge cases — and demonstrably produces the same values.
- Comments carry their weight (refused-pair wiring, the deliberate no-memoise rationale with the ~70 MB figure, the `_missing_child` endgame note) without narrating the obvious.
- Test updates are proportional: `test_discard_wiring` exercises `build_discard_game` directly (no monkeypatch), and the new round-trip test pins the 48-bit key budget.

**Summary: 0 blocking, 2 minor/robustness notes (both optional). Correctness of the RAM rewrite verified end-to-end — enumeration equivalence, child-stage ordering, off-tree recursion, and bit-identical Smoke/Six/Scotland values all confirmed independently.**

PR #41, opened 2026-07-05
Default to exact solve; k=3 as opt-in fast preview (#16)

### LINKED ISSUE #16: B5: unrestricted-solve measurement + core-language decision
The k-restriction is an approximation, not just a speed knob. After B1–B4: measure the exact (no-k) 8-player solve (~30–100× the k=4 tree). If < ~10 min in Python, default to exact with k as fast-preview. Outcome feeds the M4 core-tech spike (TypeScript port vs Rust/WASM).

Acceptance: documented timings for k=3/k=4/exact; decision recorded in PLAN.md.

svknoe:
Closes #16.

Implements the B5 decision recorded in PLAN.md (2026-07-05): the exact,
fully-unrestricted solve lands at ~3 min / <1 GB on a full 8-player team —
well under the pre-agreed ~10-min bar — so the CLI now defaults to the **true
equilibrium** rather than the k-restriction heuristic. The heuristic stays as
an explicit **fast preview**.

## Changes

- **`SolverConfig` defaults** (`drafter/solver/context.py`):
  `restrict_attackers` → `False` (exact by default; `initialise()` already maps
  this to `restriction=None` = no heuristic anywhere). `restricted_attackers_count`
  → `3`, so a bare `SolverConfig(restrict_attackers=True)` *is* the documented
  k=3 preview.
- **Startup mode prompt** (new `drafter/data/set_solve_mode.py`, wired into
  `app.run()`): after the team menu, choose **Exact** (default, ~3 min) vs
  **Fast preview** (k=3, ~30 s). The I/O is split from a pure
  `config_for_mode(preview)` so the mapping is unit-testable without a TTY.
- **Docs**: README's stale "runtime > 1 hour, change 4→3" disclaimer and
  CLAUDE.md's knob note are replaced with the exact-by-default / fast-preview
  model + the startup prompt. PLAN.md already records the decision (it's the
  direction doc, not a status tracker), so it's unchanged.
- **Tests** (new `drafter/tests/test_solve_mode.py`): default-guard tests that
  the shipped default is exact and the preview maps to k=3 restricted. The
  existing golden-value tests deliberately pin `k` explicitly, so they would
  *not* catch the default silently flipping back — hence these guards.

## Verification

- Fast suite: **92 passed, 1 deselected** (`python -m pytest`).
- CI smoke test (`scripts/smoke_draft.py` on the Smoke fixture): completes.
  Smoke is 4-player, exact at any k, so CI stays fast/green.
- Existing k-pinned golden values are **unchanged** by design.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -108,10 +108,17 @@ best when neither does (refused-vs-refused and last-players games).
   arrays (above) cut Scotland k=3 **peak RAM from ~1.8 GB to ~120 MB** (~15×);
   fresh solve ≈ 28 s (enumeration ~8 s, values ~20 s). The old 630 MB JSON disk
   cache is gone — the in-memory solve is fast and cheap enough not to need it.
-- `restricted_attackers_count` (k) in `SolverConfig` (drafter/solver/context.py) is the only current knob:
-  each select-attackers step only considers the k heuristically best
-  attackers per side (heuristic: advantage vs defender minus average vs
-  the rest). k=4 default, k=3 for quick runs.
+- **B5 (decided, issue #16):** the exact (unrestricted) solve lands at ~3 min
+  / <1 GB — under the ~10-min bar — so the CLI **defaults to exact**
+  (`SolverConfig.restrict_attackers=False`), the true equilibrium with no
+  heuristic anywhere. The k-restriction is now an opt-in **fast preview**:
+  `restrict_attackers=True` restricts each select-attackers step to the
+  `restricted_attackers_count` heuristically best attackers per side (heuristic:
+  advantage vs defender minus average vs the rest), and `restricted_attackers_count`
+  defaults to 3 (~30 s at 8 players). `python -m drafter` offers exact-vs-preview
+  as a startup prompt (drafter/data/set_solve_mode.py). The restriction saturates
+  from the bottom up (k≥3 → 4-player stage exact, k≥5 → 6-player, k≥7 → whole
+  draft exact), so k=7 ≡ exact.
 
 ## Input format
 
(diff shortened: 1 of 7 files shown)


svknoe:
## Code review (one-pass, reviewer agent)

**Verdict: no high-confidence issues — correct, consistent, well-tested. Safe to merge.**

Reviewed the full `git diff main...issue-16-exact-by-default`. Verified against the code paths and conventions:

**Correctness**
- `initialise()` maps `restrict_attackers=False` → `restriction=None`, and `get_attackers_team_permutations` skips restriction entirely when `ctx.restriction is None` — so the flipped default genuinely produces the exact solve.
- `config_for_mode(True/False)` returns exactly the two documented configs; the pure/IO split is as the docstring claims and unit-tested without a TTY.
- The `context` import removed from `app.py` had no other use — no dangling reference.

**Consistency**
- `set_solve_mode.py` mirrors `set_enemy_team.py` (InquirerPy, `# 3rd party`/`# local source`, `.format()` not f-strings) and honours "knobs live in SolverConfig; interactive prompt, not a CLI flag."
- Golden tests are genuinely unaffected — `_solve_fixture.py` always passes `restrict_attackers=True` with an explicit `k`. `test_solve_mode.py` fills the default-guard gap the golden tests explicitly disclaim.

**Docs**
- The saturation claim (k≥3 → 4-player exact, k≥5 → 6-player, k≥7 → whole draft exact ⇒ k=7 ≡ exact) is mathematically correct (attackers chosen from n−1 per stage). Runtime figures match PLAN.md B5 / CLAUDE.md B3.

**One below-threshold observation (~55), addressed:** `scripts/smoke_draft.py` builds a bare `SolverConfig()`, which now means *exact*, so a **manual** 8-player smoke run costs ~3 min instead of ~30 s. This is the intended consequence — smoke_draft deliberately exercises the shipped default — and CI is unaffected (it uses the 4-player Smoke fixture, exact at any k, ~1 s). Fixed the surprise with a docstring note in a6a1324 recording the timing and how to opt into the k=3 preview.

CI green; existing k-pinned goldens unchanged by design.

PR #42, opened 2026-07-05
#17 core-tech spike: TS engine + conformance + in-browser benchmarks -> TypeScript decided

### LINKED ISSUE #17: Core-tech spike: TypeScript port vs Rust/WASM
Port the B-shaped core (bitmask states, tiny zero-sum LPs, one backward pass) both ways far enough to benchmark an 8-player k=4 solve in the browser. Constraint from PLAN.md: fully client-side (pairing matrices are competitive secrets — nothing leaves the browser). Pyodide only as a throwaway reference point.

Acceptance: decision doc (docs/web-core-decision.md) with in-browser timings; k=4 must feel interactive (seconds, in a web worker).

svknoe:
Closes #17.

## What this is

The M4 core-tech spike: port the B-shaped engine core to TypeScript, prove it against the Python engine, benchmark it **in a browser worker** at k=3 / k=4 / exact, and make the TS-vs-Rust→WASM call.

## Result: TypeScript, decisively — no Rust→WASM

Same machine (Ryzen 9800X3D), Scotland 8×8, single-threaded both sides. TS measured in a web worker in headless Chrome behind the real §3 protocol; native Python re-measured the same day (peak working set via psutil):

| config | native Python | TS in-browser | speedup | TS worker heap (measured, retained) | TS peak alloc (engine accounting) |
|---|---|---|---|---|---|
| k=3 | 28.6 s / 111 MB | **0.42 s** | ~68× | 15 MB | 15 MB |
| k=4 (the decision bar) | 71.0 s / 286 MB | **0.81 s** | ~87× | 26 MB | 32 MB |
| exact (k=7) | 194.9 s / 801 MB | **2.02 s** | ~96× | 57 MB | 108 MB |

The bar was "k=4 feels interactive"; the measured outcome is that even the **exact solve is ~2 s in-browser**. Both native phases turn out to be interpreter-bound, not numpy-bound: enumeration (pure loops) gets ~6× from V8, and the value phase gets ~27× from replacing scipy-linprog-per-call with the closed-form 2×2 path + a bespoke ~100-line dense simplex on `Float64Array`. Full verdict + methodology written into `docs/web-design.md` §7.1 (per the goal directive, the decision lives in §7 rather than a separate `docs/web-core-decision.md`); §7's decision table and PLAN.md M4 updated. Downstream consequence recorded in §7: the web app can default to **exact**, mirroring the CLI's B5 decision.

## What's in the PR

- **`web/src/engine/`** — the TS engine core: 40-bit packed gamestate keys as plain numbers (§7.2's 53-bit trap avoided — the Python engine's fourth role nibble is unnecessary since discard states are never keyed), BFS enumeration with exact-size child buffers, the k-restriction heuristic, saddle + 2×2 closed form + bespoke simplex (Bland's rule, no LP dependency), value-only backward induction into sorted parallel `Float64Array`s, on-demand strategy recompute incl. off-tree recursion.
- **`web/src/worker/`** — the §3 worker contract: `solve` (real `progress` events) → `solved {expected, root}`, `node {path}` → `NodeResult` (per-choice `prob`/`ev` + the Why-panel payload), `reset`. The `solved` message carries an extra spike-only `stats` block for the bench.
- **`web/src/conformance/`** — fixtures exported from the Python engine (`scripts/export_conformance_fixtures.py`: Smoke exact, Six exact+k=4, Scotland k=3/k=4/exact) and the vitest suite asserting: root values to 1e-9, **exact** per-(n,stage) gamestate counts (pins enumeration order and the restriction heuristic), exact sample-node payoff matrices + option labels along an argmax walk, and both engines' strategies as ε-equilibria of the shared matrix (mixed strategies are not unique; values are — per §7.2 "conformance, not bit-equality").
- **`web/bench/` + `web/index.html`** — the in-browser benchmark harness (drives the real worker; validates every value against the fixture) plus an env-gated Node-side memory bench.
- **`.github/workflows/web-ci.yml`** — typecheck + the full conformance suite (incl. Scotland k=4/exact) on every push/PR touching `web/`.

## Verification

- `npm test` (fast) and `CONFORMANCE_SLOW=1 npm test` (full): 8 passed, incl. Scotland exact conforming to 1e-9.
- Python fast suite still green (92 passed); no engine code touched.
- Bench run end-to-end in headless Chrome via CDP and in the preview browser (numbers above; two independent browser runs agree within noise).

## Notes for reviewers

- Mixed-strategy conformance is deliberately ε-equilibrium-based, not entry-wise: HiGHS and the TS simplex may pick different vertices of a degenerate game's equilibrium polytope. Values are unique and asserted directly.
- The 2×2 discard game exists twice by design (allocation-free inlined copy in the induction hot path, labelled copy in `buildGame` for node queries); the fixtures pin both.
- `web/package-lock.json` is committed for `npm ci` in CI. `psutil` was used only ad hoc for the native measurement and is not added to `pyproject.toml`.


svknoe (commented):
## Review — one pass, findings ordered by severity

The headline first: I traced the TS engine line-by-line against the Python sources (`packing.py`, `team_permutation.py`, `utilities.py`, `games.py`, `game_state_dictionaries.py`, `strategy_dictionaries.py`) and found **no semantic divergence** in the port itself. The 40-bit key scheme is sound, the discard game (both copies) matches `build_discard_game` exactly including the issue #32 refused-return semantics, and the restriction heuristic reproduces Python's indexing, summation order, division order, ranking sign and stable-sort tie-breaks. Details in the "verified" section at the bottom. The findings below are docs-accuracy issues, one small protocol bug, and test/CI gaps.

### Medium

**M1. §7.1's per-phase speedup sentence mixes two different configs.**
"enumeration … (7.97 s → 1.26 s, ~6×) … the value phase … turns 20.6 s into 0.76 s (~27×)". But 7.97 + 20.6 = 28.57 s — that's the **native k=3** total from the same table (28.6 s), while 1.26 + 0.76 = 2.02 s is the **TS exact** total. So the "before" numbers are the native k=3 phase split and the "after" numbers are the TS exact phase split — not a like-for-like comparison, and the ~6×/~27× phase figures visibly contradict the ~96× whole-solve speedup three lines above (a reader who multiplies out the phases gets ~14×). The native exact enumeration cannot be 7.97 s (exact enumerates ~3.6M keys vs k=3's ~850k; and the phases would have to sum to 194.9 s). The verdict is unaffected — TS wins at every rung — but this is the decision-record paragraph, so the numbers should be internally consistent: either quote per-config phase splits for both engines, or label these explicitly as "the historical k=3 fresh-solve problem (28.6 s) vs the TS exact solve (2.02 s)".

**M2. PLAN.md M4 numbers disagree with §7.1.**
PLAN.md says "~1.8 s (exact) vs native Python 28/61/190 s"; §7.1's table (and the PR body) says 2.02 s exact and 28.6 / 71.0 / 194.9 s native. Looks like PLAN.md kept the old B5 ladder numbers and an earlier bench run. One of the two should be corrected (presumably PLAN.md, since §7.1 is the stated decision record).

**M3. `nodeResult` hardcodes `round: 3` for the `done` state** (`web/src/engine/engine.ts`, the `if (done)` return). For a 4×4 matrix the draft completes in round 1, for 6×6 in round 2. Should be `(this.n - 4) / 2 + 1` (the states are still at the 4-player select_attackers level when `done` is set). The Smoke walk in `nodeResult.test.ts` checks `stage`/`choices`/`why` on the done node but not `round`, so it doesn't catch this.

**M4. The off-tree recursion path is untested.**
The PR body advertises "on-demand strategy recompute incl. off-tree recursion", but nothing exercises it: every fixture sample node lies on the argmax walk of the (restricted) solved tree, and `nodeResult.test.ts` uses the exact Smoke solve where nothing is ever off-tree. `extensionValues`, the `gamestateValue` recursive fallback, and `buildDiscardGame`'s non-4-player `child()` recursion are all dead code as far as the test suite is concerned. One test that solves Scotland (or Six) at k=3 and then walks a `nodeResult` path through a deliberately non-restricted attacker pair — asserting the resulting node's value against the same engine solved exact (or a fixture value) — would cover the whole chain.

**M5. The anti-drift guard has a blind spot: web CI never fires on `drafter/**` changes.**
`web-ci.yml` triggers on `web/**` and the exporter script only. If the Python value model changes (the exact scenario §7.2 worries about), the committed fixtures go stale and web CI stays green — TS still matches the stale fixtures. Catching that automatically would mean re-exporting in CI (expensive: Scotland exact ~3 min in Python), so at minimum: document the re-export obligation where it will be seen (a comment in `web-ci.yml` + the "Bump CACHE_FORMAT_VERSION" paragraph's neighbourhood in CLAUDE.md, or PLAN.md's single-engine notes), and consider a cheap freshness marker (e.g. exporter writes the drafter git SHA into the fixture; a fast CI step fails if `drafter/**` changed since).

### Low

**L1. The simplex ratio-test tie-break is not strictly Bland's rule.**
`ratio < bestRatio + EPS && basis[i] < basis[pivotRow]` accepts a row whose ratio is up to EPS *larger* than the incumbent and then raises `bestRatio` to it, so "tie" drifts by up to EPS per acceptance; the anti-cycling guarantee the comment claims holds only for exact ties. In practice the `maxIterations` throw (surfaced as a worker `engine-error`) is the real backstop, and at ≤21×21 this will likely never fire — but the comment overstates the guarantee. Either soften the comment or make the tie-break exact (compare ratios exactly; the entering rule — first negative reduced cost — is already Bland-correct).

**L2. Simplex strategies are neither clamped nor renormalised.**
Reduced costs in (-EPS, 0] can put tiny negative probabilities into `NodeResult.prob` / `why.myStrategy` (the Python side takes `np.abs` of the marginals). The conformance suite tolerates `>= -tolerance`, but the trainer UI will eventually render these. A `max(0, ·)` + renormalise at the `solveGame` exit would be cheap and make the §3.3 payload strictly a distribution.

**L3. `worker.ts` `enumerateMs === 0` sentinel.** If the first `inducting` callback lands at a measured 0 ms (or a matrix so small the phase boundary is sub-timer-resolution), the phase split shifts to the next callback. Harmless spike instrumentation; a `number | null` sentinel is the robust form if ever touched.

**L4. Worker doesn't validate `k`.** §6 says the worker assumes a clean `Matrix`, but `k` comes over the wire too: `k <= 1` empties the attacker-pair levels and the solve dies deep inside `induct` with a confusing lookup error (caught and posted as `engine-error`, so no hang — just a bad message). A one-line guard (`k === null || k >= 2`) would give a real error code.

**L5. §7.1 "roomy margin (~8-15× under native Python's 801 MB)"** — 801/108 ≈ 7.4×, so the low end rounds up. "~7–14×" is the honest range.

**L6. Exporter reaches into a private helper.** `scripts/export_conformance_fixtures.py` uses `strategy_dictionaries._child_value_getter(ctx)`. Fine for a script, but the exporter is now load-bearing (web CI paths reference it, fixtures are the anti-drift guard) — worth either a public alias or a comment in `strategy_dictionaries.py` noting the external caller.

### Verified (so you know what was checked, not just what was flagged)

- **packing.ts / 40-bit keys:** max key < 2^40, exact in a float64; bitwise ops touch only 20-bit team codes; the key split uses division, not `>>`. Dropping the discard nibble is safe — I confirmed the TS enumeration never keys discard states (it generates the four grandchildren `{fA,fB}×{eA,eB}` directly) and that this exactly matches Python's discard→none two-step expansion in `get_discard_team_permutations` + `get_none_team_permutation`, dedupe included. No overflow, no collision.
- **zeroSum.ts vs utilities.py:** the 2×2 mixed-first-then-saddle order, the saddle tie-breaks (`>=` / `<=` picking index 0), the larger-game first-index argmax/argmin (strict `>` / `<` keeps the first, matching numpy), and the exact `maximin === minimax` equality test are all bit-faithful. The simplex solves the column player's LP — the dual of Python's row-side LP — and the dual extraction from the slack reduced costs has the correct sign (no `abs` needed, unlike scipy's marginals); the positivity shift and un-shift match (`shiftedValue + shift - 1`).
- **restriction.ts vs get_heuristically_best_attackers:** `worst[a][d]` / `best[d][a]` (transposed) / `neutral` transposed for the enemy side; field sum in ascending-opponent order then one division, matching Python's float-op sequence exactly; `ranking_sign` ∓1; stable sort on both sides (ES2019 guarantees it), so ties keep ascending player order like Python's `sorted`. The ranked (not re-sorted) order feeding `itertools.combinations`-order pair enumeration is preserved, and the fixture stage counts + sample-node labels pin it.
- **Discard game, both copies:** `discardGameValue` (hot path) and `buildDiscardGame` (labelled) have identical AA/AB/BA/BB terms, and both match `games.build_discard_game`: kept-attacker terms (`fD_eB` pairs with refusing `eA`, etc.), refused pair returning to the pools, the 4-player endgame's `fN_eN` neutral term, and the row=refused-ENEMY-attacker / col=refused-FRIENDLY-attacker label orientation.
- **Exporter argmax walk:** the refusal application is correct — `refused_enemy = friendly_options[i]` (row label), `refused_friendly = enemy_options[j]` (col label), both appended back to their own side's pool.
- **nodeResult refusal semantics:** `move.my` validated against the *enemy* attackers, `move.enemy` against the friendly ones; both refused players return to their pools — consistent with the engine, the exporter, and issue #32.
- **Enumeration:** the exact-size child-buffer assumption (uniform per-level child counts) is sound — every state on an (n, stage) level has the same remaining-popcounts and, under restriction, `min(k, remaining)` eligible attackers — and it's runtime-checked (`offset !== children.length` throws). `Float64Array.sort` is numeric; dedupe mirrors `np.unique`.
- **Induction order and child-store wiring** match `solve_order`/`child_stage`; the 4-player select_attackers stage correctly never touches its (absent) child store.
- **Worker protocol:** `solve`/`node`/`reset` requests and `progress`/`solved`/`nodeResult`/`reset-ok`/`error` responses match §3.1–§3.3 and §3.5 (enemy opening mix via `root.why.enStrategy` is there); the `stats` block is a documented spike-only extra that a §3-pure UI ignores.
- **Conformance suite design** is sound: values (unique) asserted directly at 1e-9; strategies as ε-equilibria with a justified looser bound for HiGHS output; exact stage counts and exact sample-node payoffs/labels pin enumeration order and the restriction. The main strengthening opportunity is M4 (off-tree) — the on-tree surface is well covered.
- **Speedup table arithmetic** checks out (28.6/0.42 ≈ 68, 71.0/0.81 ≈ 88, 194.9/2.02 ≈ 96), and the PR body's claims match §7.1. Repo conventions: `checkout@v7`/push-on-main-only mirrors the existing `ci.yml`, `.gitignore` additions are right, no Python engine code touched, PLAN.md updated (modulo M2).

Nice spike overall — the conformance-first approach and the exact per-stage count pinning are exactly the right guard rails, and the decision write-up in §7.1 is where it belongs.

🤖 Reviewer agent (Claude Code)

svknoe:
Thanks — sharp pass. Everything actionable is addressed in the follow-up commit; per-finding disposition:

**M1 (mixed-config phase speedups in §7.1) — fixed.** You read it exactly right: I had paired the native *k=3* phase split with the TS *exact* split. The paragraph now compares exact-to-exact using the same-day native re-measurement (enumeration 125.1 s → 1.26 s ≈ 99×, value phase 69.8 s → 0.76 s ≈ 92×), which is also consistent with the ~96× whole-solve figure in the table.

**M2 (PLAN.md numbers) — already fixed before the review landed** (41f0214 crossed with your pass): PLAN.md now quotes 28.6/71/195 s and ~2.0 s exact, matching §7.1.

**M3 (`round: 3` hardcoded on `done`) — fixed.** Now derived from the current level (`(n − 4)/2 + 1`), and the Smoke walk asserts `round === 1` on its done node.

**M4 (off-tree recursion untested) — fixed** with `web/src/engine/offTree.test.ts`, using the saturation property as the correctness anchor: for Six at k=3 every 4-player subtree is already exact, so an off-tree 6-player select_attackers node must match the exact engine's payoff and value to 1e-9 — and it does. The test also asserts the restriction actually binds (3 vs 10 pairs), that `extensionValues` memoisation kicks in and is stable, and that `nodeResult` follows the off-tree path end to end. (Small API addition: `extensionCount()`.)

**M5 (fixtures can go silently stale on `drafter/**` changes) — documented, marker deferred.** The re-export obligation now lives in both places a value-model change would be made from: CLAUDE.md (appended to the existing "bump `CACHE_FORMAT_VERSION`" rule, which is the same trigger) and a comment in `web-ci.yml` explaining why `drafter/**` is deliberately not a trigger. I'm deferring the SHA freshness marker: it has an awkward failure mode (any `drafter/**` refactor — even value-neutral — would demand a re-export or a manual override), and the single-engine end state (PLAN.md) retires the two-engine window entirely. If the transition drags, the marker is the right next step.

**L1 (fuzzy ratio-test tie-break) — fixed.** The tie-break is now exact (`ratio < best`, ties on `===` by lowest basis index) and the comment no longer overclaims: it names the iteration cap as the hard backstop.

**L2 (unclamped strategies) — fixed.** Simplex output is clamped at 0 and renormalised before leaving `solveGame`, so `NodeResult.prob`/`why.*Strategy` are always distributions. Conformance (ε-equilibrium) unaffected; full suite still green.

**L3 (`enumerateMs === 0` sentinel) — fixed** with a `number | null` sentinel.

**L4 (unvalidated `k`) — fixed.** The worker rejects `k !== null && (!Number.isInteger(k) || k < 2)` with a `bad-request` error code before constructing the engine.

**L5 ("~8–15×") — fixed** to ~7–14×.

**L6 (private `_child_value_getter`) — fixed.** The exporter now composes public API only (`strategy_dictionaries.game_value` + `game_state.get_gamestate_from_key`); re-ran the full export afterwards and the committed fixtures were byte-identical — except the `pythonSolveSeconds` timing metadata, which the byte-identity check itself exposed as re-export diff noise. Since M5 makes re-export a routine obligation, the field is now dropped entirely: exporter output is deterministic (verified — a re-export over the committed fixtures diffs empty).

Verification after all of the above: `tsc --noEmit` clean; full conformance suite (`CONFORMANCE_SLOW=1`) 11 passed, including Scotland exact at 1e-9; Python fast suite untouched and green.

Issue #21, opened 2026-07-03
Slice 3: full train-against-the-bot


svknoe:
The MVP (decided 2026-07): interactive draft mirroring the CLI loop — suggested strategies with probabilities, sampled bot moves, undo/back, final pairings with score vs expected value. Plus a 'why' panel showing the payoff matrix at the current node.

Acceptance: a full 8-player training draft can be played on the live site against the bot; outcome summary matches CLI semantics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

svknoe:
Spec for the draft-summary decomposition (agreed with the owner in design review; the report mock already shows it): the summary panel splits the total delta vs pre-draft plan into **regret** (my picks) and **variance** (reveal luck).

- Per-decision regret: at each of MY decisions, `regret = max(choices[].ev) − ev(chosen)`, using `NodeResult.choices[].ev` (EV vs the opponent's equilibrium mix at that node — docs/web-design.md §3.3). 0 = perfect pick. Bot decisions contribute nothing.
- `variance = totalDelta − Σ regrets` (attributable to which branch of its mix the bot sampled).
- Invariant to assert in tests: regrets ≥ 0 (up to LP tolerance) and the decomposition sums exactly.
- Copy rule: when Σ regrets ≈ 0, the verdict says the deficit/surplus was variance ('clean draft, unlucky reveals'), never 'the bot out-drafted you'.
- UI: worst-first list of my non-zero-regret decisions ('R2 attackers: sent Orks + T'au, best was Tyranids + Orks, −2.3').

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #45, opened 2026-07-05
Slice 2 (#20): in-browser solve + strategy tables

### LINKED ISSUE #20: Slice 2: in-browser solve + strategy tables
Run the ported core in a web worker with a progress bar; show the top-level equilibrium and per-stage strategy tables with probabilities.

Acceptance: k=4 solve of an 8×8 matrix completes in seconds on a mid-range laptop; results match the Python engine's golden values.

svknoe:
Closes #20. Wires the solve view to the real #17 §3 worker — no engine changes.

## What's here
- **`useSolve` hook** — owns one `WorkerClient` for the tab, drives `solve()` with **real** progress events (`frac`/`phase`), exposes `{status, progress, phase, result, error, solve, reset}`. Exact (`k=null`) is the default (§7); **k=3 fast preview** is opt-in.
- **`SolveView`** (§3.5) — pre-run (heading + description + exact/preview toggle + Run), solving (live progress bar), done (the **expected team result** as `80 ± margin` → e.g. `86 – 74` + verdict + "sum of 8 games · 160 points total"), and **both opening-defender mixes** (mine from `root.choices`, enemy from `root.why.colLabels`/`enStrategy`). The mode toggle is reachable from the results screen too, so k=3 is switchable post-solve.
- **`StrategyTable` + `ProgressBar`** (reused by #21); `scale.teamResult` (TDD).
- **App**: enables the Solve tab, invalidates a solved result on any matrix edit (§3 reset-on-change).
- Moves shared CSS utilities to `theme.css`.

## Verification
- `typecheck` clean; `npm test` **64 passed / 3 skipped** (full `CONFORMANCE_SLOW=1` run passes — engine values unchanged, pinned to 1e-9).
- **In-browser, real worker** (Scotland sample): exact → **86 – 74, "favored by 6"** (golden `expected=6.137780`); k=3 preview → **86 – 74** labelled "fast preview (k=3)" (golden `5.875946`); each completes in ~seconds. Opening mixes are true equilibria (mine: Mariusz 86% / Petter 14%; enemy: Drukhari 66% / Sisters 34%) — correctly different from the mockup's throwaway stub (§5). No console errors.

Note: preview-tool screenshots were timing out this session (a tooling quirk, not an app issue — the page responds to all DOM queries with zero console errors), so verification evidence is DOM-level rather than an image.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,21 +1,34 @@
 import { useEffect, useMemo, useState } from 'react';
 import { MatrixEditor } from './components/MatrixEditor';
+import { SolveView } from './components/SolveView';
 import { blank } from './model/matrix';
 import type { EditorMatrix } from './model/matrix';
 import { loadState, saveState } from './model/storage';
 import type { AppState, Settings } from './model/storage';
+import { validateMatrix } from './model/validation';
+import { useSolve } from './worker/useSolve';
 
 type Screen = 'editor' | 'solve' | 'trainer' | 'summary';
 
 export function App() {
   const [state, setState] = useState<AppState>(() => loadState());
   const [screen, setScreen] = useState<Screen>('editor');
+  const [k, setK] = useState<number | null>(null); // null = exact (§7 default)
+  const solve = useSolve();
 
   // Persist the whole blob whenever it changes (§4.2 auto-save).
   useEffect(() => saveState(state), [state]);
 
   const matrix = useMemo(() => state.current ?? blank(8), [state.current]);
   const { settings, saves } = state;
+  const solvable = useMemo(() => validateMatrix(matrix).ok, [matrix]);
+
+  // A matrix edit invalidates any solved result (§3: reset on matrix change).
+  useEffect(() => {
+    if (solve.status !== 'idle') solve.reset();
+    // Only react to matrix identity; solve is stable enough for this guard.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [matrix]);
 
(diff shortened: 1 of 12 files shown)


svknoe:
## Reviewer-agent pass

A code-reviewer subagent reviewed the branch diff and confirmed **engine/worker unchanged** (`git diff main...HEAD -- web/src/engine web/src/worker/worker.ts` empty). It verified as correct: the `teamResult` 80±margin math (always sums to 160), the enemy-mix zip (`colLabels[j]`/`enStrategy[j]` aligned), null-`why` handling, the `[string,string]` name case, lazy worker creation (jsdom-safe), the reset-on-edit effect deps (no loop), and the CSS-utility relocation (clean move, no editor regression).

**One Important finding (confirmed):** a stale-result race in `useSolve`. Editing the matrix during an in-flight solve calls `reset()`, but a posted worker request can't be cancelled (the engine runs synchronously in its thread), so the old solve's `.then` would land a stale `done` result *after* the reset — showing matrix A's equilibrium while `state.current` is now B.

### Resolution — f0730b0
- Added a generation token to `useSolve`, bumped on every `solve()` and `reset()`; the progress/then/catch handlers no-op when superseded.
- Regression test drives the exact race with a deferred-response fake worker (RED before the guard: `expected 'done' to be 'idle'` → GREEN after).
- `typecheck` clean; `npm test` 65 passed / 3 skipped.

Minor note (not fixed, by design): flipping the exact/preview toggle on the results card updates the caption before **Re-run** is pressed — Re-run is right there, so the momentary mislabel is acceptable.

